PR #7, opened 2019-07-10
DEV-821 Add option to specify CMEK

pauldwolfe:
Passed in as an argument, when present all runtime buckets will set the
kms property on creation (note will not change old buckets already using
Google keys).

Also necessary to increase storage timeouts and use the .getResult() to
ensure the the entire copy operation completes.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/storage/StorageProvider.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/storage/StorageProvider.java
@@ -21,8 +21,8 @@ public class StorageProvider {
         return builder.setCredentials(credentials)
                 .setProjectId(arguments.project())
                 .setTransportOptions(HttpTransportOptions.newBuilder()
-                        .setConnectTimeout(Integer.MAX_VALUE)
-                        .setReadTimeout(Integer.MAX_VALUE)
+                        .setConnectTimeout(0)
+                        .setReadTimeout(0)
                         .build())
                 .build()
                 .getService();
--- a/cluster/src/main/java/com/hartwig/pipeline/storage/StorageProvider.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/storage/StorageProvider.java
@@ -1,6 +1,7 @@
 package com.hartwig.pipeline.storage;
 
 import com.google.auth.oauth2.GoogleCredentials;
+import com.google.cloud.http.HttpTransportOptions;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageOptions;
 import com.hartwig.pipeline.Arguments;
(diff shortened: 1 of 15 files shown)


cluster/src/test/java/com/hartwig/pipeline/storage/RuntimeBucketTest.java
```diff
@@ -133,6 +133,19 @@ public void namespacesListOperationPrefix() {
         assertThat(listOptionCaptor.getValue()).isEqualTo(Storage.BlobListOption.prefix("test/prefix"));
     }
 
+    @Test
+    public void usesCmekWhenProvided() {
+        RuntimeBucket.from(storage, NAMESPACE, referenceRunMetadata(), Arguments.testDefaultsBuilder().cmek("key").build());
+        assertThat(bucketInfo.getValue().getDefaultKmsKeyName()).isEqualTo(
+                "projects/hmf-pipeline-development/locations/europe-west4/keyRings/hmf-pipeline-development/cryptoKeys/key");
```
nedleitch: Do we want to parameterise this? Makes the test somewhat brittle if we don't.

cluster/src/main/java/com/hartwig/pipeline/storage/StorageProvider.java
```diff
@@ -16,7 +17,15 @@ private StorageProvider(final Arguments arguments, final GoogleCredentials crede
     }
 
     public Storage get() {
-        return StorageOptions.newBuilder().setCredentials(credentials).setProjectId(arguments.project()).build().getService();
+        StorageOptions.Builder builder = StorageOptions.newBuilder();
+        return builder.setCredentials(credentials)
+                .setProjectId(arguments.project())
+                .setTransportOptions(HttpTransportOptions.newBuilder()
+                        .setConnectTimeout(Integer.MAX_VALUE)
+                        .setReadTimeout(Integer.MAX_VALUE)
```
nedleitch: I think you can pass "0" here if you want an infinite timeout.

[reference](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java)

pauldwolfe: Good idea

cluster/src/test/java/com/hartwig/pipeline/storage/RuntimeBucketTest.java
pauldwolfe: Struggle a bit with this, as if I parameterize it, I basically end up rewriting the code I wrote in the RuntimeBucket itself. But maybe that's still better than fixing a million littles tests if the Arguments.testDefaults() change. I'll implement your suggestion

PR #8, opened 2019-07-12
DEV-804 Refactor of GRIDSS subsystem

nedleitch:
There are some first tentative steps to a more functional way of doing things inline with what's been done in the somatic caller. See `PreprocessFunctional.java`.


cluster/src/main/java/com/hartwig/pipeline/calling/structural/gridss/command/CollectGridssMetrics.java
```diff
@@ -32,10 +33,8 @@ public String arguments() {
                         .asBash();
     }
 
-    // This GRIDSS command uses the "-o" argument to mean the fully-qualified basename for the output files. It will
-    // generate filenames for each of the requested metrics based upon that starting point.
     public String outputBaseFilename() {
-        return VmDirectories.outputFile(format("%s_metrics", GridssCommon.basenameNoExtensions(inputBam)));
+        return format("%s/%s", workingDirectory, new File(inputBam).getName());
```
pauldwolfe: do we need to use the input bam or can we just use the sample name for this file? I find it a little awkward some files are named "CPCT12345678T.sorted.something", would be nice to drop the sorted.

cluster/src/main/java/com/hartwig/pipeline/calling/structural/StructuralCaller.java
```diff
@@ -82,7 +83,8 @@ public StructuralCallerOutput run(final SomaticRunMetadata metadata, final Align
         InputDownload referenceBam = new InputDownload(pair.reference().finalBamLocation());
         InputDownload referenceBai = new InputDownload(pair.reference().finalBaiLocation());
 
-        bash.addCommands(asList(tumorBam, referenceBam, tumorBai, referenceBai, referenceGenomeDownload, gridssConfigFiles, gridssPonFiles));
+        bash.addCommand(new BatchInputDownload(referenceBam, referenceBai, tumorBam, tumorBai));
```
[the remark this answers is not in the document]
nedleitch: Yeah it was an optimisation that I think is applicable across the board. Kinda got lumped in with the grab-bag of GRIDSS stuff because it was already in my workspace but yes it should have been done separately.

cluster/src/main/java/com/hartwig/pipeline/calling/structural/gridss/command/CollectGridssMetrics.java
nedleitch: Think we might be able to do away with it altogether when we make the whole thing more in the functional style.

PR #11, opened 2019-07-12
DEV-829 Add weird hacky looking thing before filter in GRIDSS

pauldwolfe:
See ticket for the rationale. Copied verbatim from P4.


cluster/src/main/java/com/hartwig/pipeline/calling/structural/gridss/stage/Filter.java
```diff
@@ -39,6 +41,11 @@ public FilterResult initialise(final String originalVcf, final String tumorSampl
 
         List<BashCommand> commands = new ArrayList<>();
         commands.add(() -> format("gunzip -kd %s", originalVcf));
+        commands.add(() -> format(
```
nedleitch: Yikes.

PR #17, opened 2019-07-18
DEV-839 Add QC_FAILED status for health checker

pauldwolfe:
This allows us to mark a run as failed for the SBP API call without
skipping cleanups as we do on a tech failure.

Also added a status check in PipelineMain to exit 1 on tech failures
from remote jobs (vms and dataproc).

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/PipelineMain.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/PipelineMain.java
@@ -11,6 +11,7 @@ import com.hartwig.pipeline.calling.somatic.SomaticCallerProvider;
 import com.hartwig.pipeline.calling.structural.StructuralCallerProvider;
 import com.hartwig.pipeline.cleanup.CleanupProvider;
 import com.hartwig.pipeline.credentials.CredentialProvider;
+import com.hartwig.pipeline.execution.PipelineStatus;
 import com.hartwig.pipeline.execution.vm.ComputeEngine;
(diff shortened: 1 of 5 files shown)


cluster/src/main/java/com/hartwig/pipeline/PipelineMain.java
```diff
@@ -58,27 +60,30 @@ public static void main(String[] args) {
...
+                            AmberProvider.from(arguments, credentials, storage).get(),
+                            CobaltProvider.from(arguments, credentials, storage).get(),
+                            SomaticCallerProvider.from(arguments, credentials, storage).get(),
+                            StructuralCallerProvider.from(arguments, credentials, storage).get(),
+                            PurpleProvider.from(arguments, credentials, storage).get(),
+                            HealthCheckerProvider.from(arguments, credentials, storage).get(),
+                            Executors.newCachedThreadPool()).run();
                     LOGGER.info("Somatic pipeline is complete with status [{}]. Stages run were [{}]", state.status(), state);
                 }
+                if (state.status() == PipelineStatus.SUCCESS) {
```
nedleitch: We could consider adding exit values to the `PipelineStatus` enum so we can just call `System.exit(state.status().code())` or something similar.

pauldwolfe: See your point, would remove the need for the if, and look pretty tidy... but I think exits codes are more in the "domain" of PipelineMain than the status itself. Going to leave it as is for now.

PR #18, opened 2019-07-18
DEV-840 Wrap up two single sample and somatic in one pipeline

pauldwolfe:
Adds a FullPipeline which will run the two single samples in parallel
and then the somatic.

The sample ids are inferred from a set id passed, so currently only
works with samples in the CPCT12345678R/T format.


cluster/src/main/java/com/hartwig/pipeline/FullPipeline.java
```diff
@@ -0,0 +1,38 @@
...
+        this.somaticPipeline = somaticPipeline;
+        this.executorService = executorService;
+    }
+
+    public PipelineState run() {
+        Future<PipelineState> referenceStateFuture = executorService.submit(referencePipeline::run);
+        Future<PipelineState> tumorStateFuture = executorService.submit(tumorPipeline::run);
+        try {
+            PipelineState singleSampleState =
+                    new PipelineState().combineWith(referenceStateFuture.get()).combineWith(tumorStateFuture.get());
```
nedleitch: Is there are reason you can't do just `referenceStateFuture.get().combineWith(tumorStateFuture.get());`, without the initial `new` object?

cluster/src/test/java/com/hartwig/pipeline/FullPipelineTest.java
```diff
@@ -0,0 +1,77 @@
...
+        when(somatic.run()).thenReturn(failed());
+        assertThat(victim.run().status()).isEqualTo(PipelineStatus.FAILED);
+    }
+
+    private static PipelineState succeeded() {
+        return new PipelineState();
+    }
+
+    private static PipelineState failed() {
+        PipelineState failedState = succeeded();
```
nedleitch: Not sure the few characters saved there are worth the confusion...

pauldwolfe: Ha Intellij's decision, will change

cluster/src/main/java/com/hartwig/pipeline/FullPipeline.java
pauldwolfe: we could but thought it was a bit arbitrary to use either the reference or tumor... will think about it

PR #22, opened 2019-07-25
DEV-847 Copy startup script to output bucket

nedleitch:
Useful for debugging, manual re-running and auditing.

The script name is shortened when it is copied to the final output bucket.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/report/StartupScriptComponent.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/report/StartupScriptComponent.java
@@ -19,6 +19,6 @@ public class StartupScriptComponent implements ReportComponent {
     @Override
     public void addToReport(final Storage storage, final Bucket reportBucket, final String setName) {
         runtimeBucket.copyOutOf("copy_of_startup_script_used_for_this_run.sh", reportBucket.getName(),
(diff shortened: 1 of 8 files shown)


cluster/src/main/java/com/hartwig/pipeline/report/StartupScriptComponent.java
```diff
@@ -0,0 +1,24 @@
...
+    public StartupScriptComponent(final RuntimeBucket runtimeBucket, final String namespace, final Folder folder) {
+        this.runtimeBucket = runtimeBucket;
+        this.namespace = namespace;
+        this.folder = folder;
+    }
+
+    @Override
+    public void addToReport(final Storage storage, final Bucket reportBucket, final String setName) {
+        runtimeBucket.copyOutOf("copy_of_startup_script_used_for_this_run.sh", reportBucket.getName(),
+                String.format("%s/%s%s/%s", setName, folder.name(), namespace, "bootstrap.sh"));
```
pauldwolfe: I think `startup.sh` is probably a better name given the history of "bootstrap"

nedleitch: Changed to `run.sh` per our conversation.

PR #23, opened 2019-07-25
DEV-846 Refactor GRIDSS subsystem

nedleitch:
GRIDSS was the odd duck of the callers as it had not been written to
conform to the SubStage setup that was used elsewhere. It has now been
rewritten to mostly look like the others.

Also all the BashCommand subclasses have been modified to use constructor
injection as is used elsewhere rather than taking parameters via method
calls. This was necessary to take advantage of the SubStage structure and
has the nice benefit of simplifying testing. The CommandFactory and
related classes have been removed.

Also some tests have been added for the StructuralCaller though there are
more to write still.

Not quite finished yet because I have not yet written a test for the
structural caller. That will be done before this branch is pushed.

Other changes were to make it so the PipeCommand always wraps its elements
in a subshell. A later commit to this branch will ensure all existing
usages do not double up on the subshells as they have often been used
together.


cluster/src/main/java/com/hartwig/pipeline/calling/structural/StructuralCaller.java
```diff
@@ -88,69 +95,46 @@ public StructuralCallerOutput run(final SomaticRunMetadata metadata, final Align
...
-        bash.addCommand(new MkDirCommand(gridssWorkingDirForReferenceBam));
-        bash.addCommand(new MkDirCommand(gridssWorkingDirForTumorBam));
+        bash.addCommand(new MkDirCommand(referenceWorkingDir));
+        bash.addCommand(new MkDirCommand(tumorWorkingDir));
 
         String preprocessSvOutputReferenceBam =
-                format("%s/%s.sv.bam", gridssWorkingDirForReferenceBam, basename(referenceBam.getLocalTargetPath()));
-        String preprocessSvOutputTumorBam = format("%s/%s.sv.bam", gridssWorkingDirForTumorBam, basename(tumorBam.getLocalTargetPath()));
+                format("%s/%s.sv.bam", referenceWorkingDir, basename(referenceBam.getLocalTargetPath()));
+        String preprocessSvOutputTumorBam = format("%s/%s.sv.bam", tumorWorkingDir, basename(tumorBam.getLocalTargetPath()));
```
pauldwolfe: Would it be possible to have this returned as output from the preprocess substage?

cluster/src/main/java/com/hartwig/pipeline/calling/structural/StructuralCaller.java
```diff
@@ -88,69 +95,46 @@ public StructuralCallerOutput run(final SomaticRunMetadata metadata, final Align
...
-        bash.addCommands(preprocessedRefSample.commands())
-                .addCommands(preprocessedTumorSample.commands())
-                .addCommands(assemblyResult.commands())
-                .addCommand(calling)
-                .addCommands(annotationResult.commands())
-                .addCommands(filterResult.commands());
 
+        String refBamPath = referenceBam.getLocalTargetPath();
+        String tumorBamPath = tumorBam.getLocalTargetPath();
+        new Preprocess(refBamPath, referenceWorkingDir, referenceSampleName, referenceGenomePath, preprocessSvOutputReferenceBam)
```
pauldwolfe: We're not using the output of the preprocess for the following stages?

cluster/src/main/java/com/hartwig/pipeline/calling/structural/StructuralCaller.java
```diff
@@ -88,69 +95,46 @@ public StructuralCallerOutput run(final SomaticRunMetadata metadata, final Align
...
+        new Preprocess(tumorBamPath, tumorWorkingDir, tumorSampleName, referenceGenomePath, preprocessSvOutputTumorBam)
+                .apply(SubStageInputOutput.of(tumorSampleName, OutputFile.empty(), bash));
+
+        Assemble assemble = new Assemble(refBamPath, tumorBamPath, jointName, referenceGenomePath, configurationFile, blacklist);
+        SubStageInputOutput result =
+                assemble.andThen(new Calling(refBamPath, tumorBamPath, "", referenceGenomePath, configurationFile, blacklist))
+                        .andThen(new Annotation(referenceBam.getLocalTargetPath(),
+                                tumorBam.getLocalTargetPath(),
+                                assemble.completedBam(),
+                                "",
```
pauldwolfe: empty string?

cluster/src/main/java/com/hartwig/pipeline/calling/structural/StructuralCaller.java
```diff
@@ -88,69 +95,46 @@ public StructuralCallerOutput run(final SomaticRunMetadata metadata, final Align
...
+                                referenceGenomePath,
+                                jointName,
+                                configurationFile,
+                                blacklist))
+                        .apply(SubStageInputOutput.of(jointName, OutputFile.empty(), bash));
+        Filter.FilterResult filterResult = new Filter().initialise(result.outputFile().path(), tumorSampleName);
```
pauldwolfe: Let's make this a function as well and pass in the final output file names, not perfect but then we can keep it all consistent (and think about returning multiple files later)

cluster/src/main/java/com/hartwig/pipeline/calling/structural/gridss/stage/Annotation.java
```diff
@@ -3,44 +3,49 @@
...
-                commandFactory.buildAnnotateUntemplatedSequence(variants.resultantVcf(), referenceGenome, tumorSampleName);
-        BgzipCommand bgzip = commandFactory.buildBgzipCommand(untemplated.resultantVcf());
-        String finalOutputPath = format("%s.gz", untemplated.resultantVcf());
-        TabixCommand tabix = commandFactory.buildTabixCommand(finalOutputPath);
-
-        return ImmutableAnnotationResult.builder()
-                .annotatedVcf(finalOutputPath)
-                .commands(asList(variants, untemplated, bgzip, tabix))
-                .build();
+        String untemplatedOutputVcf = VmDirectories.outputFile(jointName + "_untemplated.vcf");
```
pauldwolfe: Could inline some of this I think (eg bash.addCommand(new AnnotateUntemplatedSequence()).addCommand(new BgZipCommand). Might make it read slightly easier.

cluster/src/main/java/com/hartwig/pipeline/calling/structural/gridss/stage/Calling.java
```diff
@@ -0,0 +1,33 @@
...
+        this.referenceGenomePath = referenceGenomePath;
+        this.configurationFile = configurationFile;
+        this.blacklist = blacklist;
+    }
+
+    @Override
+    public BashStartupScript bash(OutputFile input, OutputFile output, BashStartupScript bash) {
+        bash.addCommand(new IdentifyVariants(referenceBam, tumorBam, input.path(), output.path(), referenceGenomePath,
+                configurationFile, blacklist));
+        return bash;
```
pauldwolfe: inline

cluster/src/test/java/com/hartwig/pipeline/calling/structural/gridss/stage/GridssSubStageTest.java
```diff
@@ -0,0 +1,29 @@
...
+import com.hartwig.pipeline.calling.SubStageTest;
+import com.hartwig.pipeline.calling.structural.gridss.CommonEntities;
+import com.hartwig.pipeline.execution.vm.BashCommand;
+
+import org.mockito.ArgumentCaptor;
+
+public abstract class GridssSubStageTest extends SubStageTest implements CommonEntities {
+    ArgumentCaptor<BashCommand> captor;
+
+    void assertBashContains(final BashCommand model) {
```
pauldwolfe: prefer not to use a class hierarchy to share code, maybe extract into BashAssertions class? Might be handy elsewhere.

cluster/src/main/java/com/hartwig/pipeline/calling/structural/StructuralCaller.java
```diff
@@ -99,68 +99,66 @@ public StructuralCallerOutput run(final SomaticRunMetadata metadata, final Align
...
         String tumorBamPath = tumorBam.getLocalTargetPath();
-        new Preprocess(refBamPath, referenceWorkingDir, referenceSampleName, referenceGenomePath, preprocessSvOutputReferenceBam)
+        new Preprocess(refBamPath, referenceWorkingDir, referenceSampleName, referenceGenomePath)
                 .apply(SubStageInputOutput.of(referenceSampleName, OutputFile.empty(), bash));
-        new Preprocess(tumorBamPath, tumorWorkingDir, tumorSampleName, referenceGenomePath, preprocessSvOutputTumorBam)
+        new Preprocess(tumorBamPath, tumorWorkingDir, tumorSampleName, referenceGenomePath)
                 .apply(SubStageInputOutput.of(tumorSampleName, OutputFile.empty(), bash));
 
         Assemble assemble = new Assemble(refBamPath, tumorBamPath, jointName, referenceGenomePath, configurationFile, blacklist);
+        String filteredVcfBasename = VmDirectories.outputFile(format("%s.gridss.somatic.vcf", tumorSampleName));
```
pauldwolfe: I'm ok with this for now as the framework does not really have a better way. Let's let it soak and see if we come up with something more clever in the future.

cluster/src/main/java/com/hartwig/pipeline/calling/structural/StructuralCaller.java
```diff
@@ -99,68 +99,66 @@ public StructuralCallerOutput run(final SomaticRunMetadata metadata, final Align
...
-                format("%s/%s.sv.bam", referenceWorkingDir, basename(referenceBam.getLocalTargetPath()));
-        String preprocessSvOutputTumorBam = format("%s/%s.sv.bam", tumorWorkingDir, basename(tumorBam.getLocalTargetPath()));
-
         String configurationFile = gridssConfigFiles.find("properties");
         String blacklist = gridssConfigFiles.find("bed");
 
         String refBamPath = referenceBam.getLocalTargetPath();
         String tumorBamPath = tumorBam.getLocalTargetPath();
-        new Preprocess(refBamPath, referenceWorkingDir, referenceSampleName, referenceGenomePath, preprocessSvOutputReferenceBam)
+        new Preprocess(refBamPath, referenceWorkingDir, referenceSampleName, referenceGenomePath)
```
pauldwolfe: Still think we should be doing something with the return value of these two

cluster/src/main/java/com/hartwig/pipeline/calling/structural/StructuralCaller.java
```diff
@@ -88,95 +95,70 @@ public StructuralCallerOutput run(final SomaticRunMetadata metadata, final Align
...
+        String refBamPath = referenceBam.getLocalTargetPath();
+        String tumorBamPath = tumorBam.getLocalTargetPath();
+        new Preprocess(refBamPath, referenceWorkingDir, referenceSampleName, referenceGenomePath)
+                .apply(SubStageInputOutput.of(referenceSampleName, OutputFile.empty(), bash));
+        new Preprocess(tumorBamPath, tumorWorkingDir, tumorSampleName, referenceGenomePath)
+                .apply(SubStageInputOutput.of(tumorSampleName, OutputFile.empty(), bash));
+
+        Assemble assemble = new Assemble(refBamPath, tumorBamPath, jointName, referenceGenomePath, configurationFile, blacklist);
+        String filteredVcfBasename = VmDirectories.outputFile(format("%s.gridss.somatic.vcf", tumorSampleName));
+        String fullVcfBasename = VmDirectories.outputFile(format("%s.gridss.somatic.full.vcf", tumorSampleName));
```
pauldwolfe: We can use OutputFile.of() here instead to reduce some duplication.

cluster/src/main/java/com/hartwig/pipeline/calling/structural/StructuralCaller.java
nedleitch: Yeah I agree. Hence my interest in running an A/B against the `master` version. I think something must have got missed somewhere. Suspect there may be more to come here.

pauldwolfe: Ok!

PR #25, opened 2019-07-26
DEV-851 Retry delete if the job still exists

pauldwolfe:
NOTE: Another commit was accidentally coupled to this one, please ignore things in performance improvements and cleanups (will be merged to master before this PR is merged)

Grasping at straws a little as we cannot duplicate this issue with some
jobs not getting deleted. Assuming that its a Google issue, this commit
will retry the delete if the job still exists.

Using failsafe again to implement the retry mech. The failure condition
is a non-null return value of jobs.get. On failed attempt we try to
delete the job again.

Annoyingly, the google api returns a 404 exception instead of a response
object when get doesn't find anything, so we have to catch that and
return null to trigger the "success" scenario.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/cleanup/Cleanup.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/cleanup/Cleanup.java
@@ -26,6 +26,7 @@ public class Cleanup {
 
     private final static Logger LOGGER = LoggerFactory.getLogger(Cleanup.class);
     private static final int MAX_RETRIES = 5;
+    private static final int RETRY_DELAY = 3;
     private final Storage storage;
     private final Arguments arguments;
     private final Dataproc dataproc;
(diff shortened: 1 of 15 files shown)


cluster/src/main/java/com/hartwig/pipeline/cleanup/Cleanup.java
```diff
@@ -49,14 +57,40 @@ public void run(SomaticRunMetadata metadata) {
                         .getJobId()
                         .startsWith(tumorRun.id())) {
                     LOGGER.debug("Deleting complete job [{}]", job.getReference().getJobId());
-                    jobs.delete(arguments.project(), arguments.region(), job.getReference().getJobId()).execute();
+                    deleteJob(jobs, job);
+
+                    Job existingJob = Failsafe.with(new RetryPolicy<>().handleResultIf(Objects::nonNull)
+                            .onFailedAttempt(objectExecutionCompletedEvent -> deleteJob(jobs, job))
+                            .withDelay(Duration.ofSeconds(3))
```
nedleitch: Strikes me as a bit odd not to have a constant for the delay if the max retries gets one.

PR #28, opened 2019-07-30
DEV-854 Simple bcl2fastq application using compute engine

pauldwolfe:
First pass, new application which runs bcl2fastq. Functionally it does
the following:
- Copy down the bcl file by the flowcell id
- Run bcl2fastq
- Copy all fastq back to a runtime bucket for the flowcell

In the future it'll have to distribute the different patients to their
respective runtime buckets.

Some awkward reuse of arguments, in the end we'll either have to
continue with a monilith argument class or extract the common ones into
a shared interface. Probably a few things we'll want to make more
shareable package-wise (or maybe separate modules).

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/execution/vm/InputDownload.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/execution/vm/InputDownload.java
@@ -25,7 +25,7 @@ public class InputDownload implements BashCommand {
 
     @Override
     public String asBash() {
-        return format("gsutil -qm cp -n gs://%s/%s%s %s%s",
+        return format("gsutil -qm cp -r -n gs://%s/%s%s %s%s",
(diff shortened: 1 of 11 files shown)


cluster/src/main/java/com/hartwig/pipeline/execution/vm/InputDownload.java
```diff
@@ -25,7 +25,7 @@ private static String localPath(final GoogleStorageLocation sourceLocation) {
 
     @Override
     public String asBash() {
-        return format("gsutil -qm cp -n gs://%s/%s%s %s%s",
+        return format("gsutil -qm cp -r -n gs://%s/%s%s %s%s",
```
nedleitch: We actually probably want to do this instead of what is currently on line 31, but wrapped in the same tertiary.

PR #33, opened 2019-08-08
DEV-868 Support single sample sets in somatic pipeline

pauldwolfe:
As it fits better within SBP automation, we've decided to do the single
sample file transfer and cleanup by skipping the somatic algos in the
somatic pipeline when the SBP API returns no tumor sample in the set.

Requires changes to the somatic pipeline to skip the steps, the
cleanup to not attempt to cleanup a non-existent tumor run, and the
api interaction to allow for a null tumor.


cluster/src/test/resources/sbp_api/get_run_no_tumor.json
```diff
@@ -0,0 +1,30 @@
+{
+  "priority": 100,
+  "status": "Validated",
+  "ini": "Somatic.ini",
```
thoutenbos: This would be SingleSample.ini

cluster/src/test/resources/sbp_api/get_run_no_tumor.json
```diff
@@ -0,0 +1,30 @@
...
+    "entity_id": 122,
+    "id": 7141
+  },
+  "cluster": "object02",
+  "ref_sample": "CPCT02290012R",
+  "slaStart": null,
+  "name": "170724_HMFregCPCT_FR13999246_FR13999144_CPCT02290012",
+  "bucket": "hmf_output_2019-27",
+  "sbp_id": 7048,
+  "tumor_sample": "CPCT02290012T",
```
thoutenbos: this would be null

PR #39, opened 2019-08-13
Bugfix 51918 Per-job status

nedleitch:


nedleitch:
Abandoning this, see PR 40 instead.

PR #40, opened 2019-08-13
Write separate completion flags for each job

pauldwolfe:


bam/src/main/java/com/hartwig/bam/after/BamIndexPipeline.java
```diff
@@ -31,24 +33,31 @@ private BamIndexPipeline(final FileSystem fileSystem, final String bamFolder, fi
         this.sortAndIndexer = sortAndIndexer;
     }
 
-    public void execute(Sample sample) throws IOException, InterruptedException {
-        String bamFileLocation = String.format("/%s/%s.bam", sourceBamDirectory, sample.name()).substring(1);
-        String unsortedBam = Bams.name(sample, localBamDirectory, "unsorted");
+    public void execute(Sample sample, StatusReporter statusReporter) {
```
pauldwolfe: Why pass the reporter as an argument here when we use it as a field in the other two?

cluster/src/main/java/com/hartwig/pipeline/execution/dataproc/GoogleDataproc.java
```diff
@@ -109,9 +110,15 @@ private String jobIdAndClusterName(final RuntimeBucket runtimeBucket, final Spar
         return untrimmed.substring(0, Math.min(untrimmed.length(), 50));
     }
 
-    private boolean isDone(final String jobName, final RuntimeBucket runtimeBucket) {
-        List<StatusCheck.Status> finalStatuses = asList(StatusCheck.Status.SUCCESS, StatusCheck.Status.FAILED);
-        return finalStatuses.contains(getStatus(jobName, runtimeBucket));
+    private boolean isDone(final Job job, final String jobName, final RuntimeBucket runtimeBucket) {
+        LOGGER.info("Checking if job is done");
+        String state = job.getStatus().getState();
+        if (state.equals("DONE")) {
```
pauldwolfe: Not sure this naming really correct. Isn't a"CANCELLED" or "ERROR" job "done"? Maybe isComplete is a better name?

cluster/src/main/java/com/hartwig/pipeline/execution/dataproc/GoogleDataproc.java
```diff
@@ -221,7 +228,7 @@ private void waitForOperationComplete(Operation operation) throws IOException {
         boolean operationComplete = isDone.test(operation);
         while (!operationComplete) {
             sleep();
-            LOGGER.debug("Operation [{}] not complete, waiting...", description.apply(operation));
+            LOGGER.info("Operation [{}] not complete, waiting...", description.apply(operation));
```
pauldwolfe: did you mean to commit this?

nedleitch: I did not. Next commit will revert it.

PR #42, opened 2019-08-14
51918 Bug Fixes

pauldwolfe:
- DEV-873
- DEV-867
- DEV-869
- Use Dataproc job output rather than the jobs api.


cluster/src/main/java/com/hartwig/pipeline/cleanup/Cleanup.java
```diff
@@ -30,50 +32,55 @@
...
+    private void cleanupSample(final SingleSampleRunMetadata metadata) {
+        if (!somaticMetadataApi.hasDependencies(metadata.sampleName())) {
+            Run run = Run.from(metadata.sampleId(), arguments);
+            deleteBucket(run.id());
+            try {
+                Dataproc.Projects.Regions.Jobs jobs = dataproc.projects().regions().jobs();
+                ListJobsResponse execute = jobs.list(arguments.project(), arguments.region()).execute();
+                for (Job job : execute.getJobs()) {
+                    if (job.getReference().getJobId().startsWith(run.id())) {
+                        LOGGER.debug("Deleting complete job [{}]", job.getReference().getJobId());
```
nedleitch: Feels like this might be more of an `INFO` message to me. And maybe the `ensureDeleted` could be called by the `deleteJob`?

PR #48, opened 2019-08-23
DEV-860 Use SSD local scratch disks for /data

nedleitch:
The pricing for local "scratch" SSDs is quite a bit
less than that of persistent SSDs, or even persistent
magnetic disks. This is predominantly due to the
ephemeral nature of these disks, but that suits our
workload perfectly.

The management is not as seamless for the local SSDs.
They are exposed as multiple devices so this commit
includes code to make a RAID-0 device out of them
(we need more than one as they are 375GB each) and
mount it to `/data`, and create some of the directories
that we used to put in the imaging script.

Further I have moved the tools out of `/data/tools`
and into `/opt/tools` as it is available at image
creation time, which is when we copy down all the
tools to the image.

The necessary code has been added to the compute
engine classes and the root disk image size dropped
to 10GB because all the space consumption for these
machines is on /data.

Note that the changes to the image introduced by
this commit make the image incompatible with any
version of the code that continues to use
`/data/tools` instead of the new `/opt/tools`.


cluster/src/main/java/com/hartwig/pipeline/Arguments.java
```diff
@@ -32,6 +32,8 @@
 
     boolean usePreemptibleVms();
 
+    boolean useScratchSsds();
```
pauldwolfe: out of curiosity, what does "scratch" mean exactly? we've been referring to them as local ssds (as do the options in the gui). Does scratch somehow describe what we're doing more accurately?

cluster/src/main/java/com/hartwig/pipeline/execution/vm/BashStartupScript.java
```diff
@@ -1,39 +1,59 @@
...
 import java.util.ArrayList;
 import java.util.List;
 
+import com.hartwig.pipeline.execution.vm.storage.StorageStrategy;
+
 public class BashStartupScript {
     static final String JOB_SUCCEEDED_FLAG = "JOB_SUCCESS";
     static final String JOB_FAILED_FLAG = "JOB_FAILURE";
-    private static final String LOG_FILE = VmDirectories.OUTPUT + "/run.log";
+    private static final String LOG_FILE = "/tmp/run.log";
```
pauldwolfe: don't love this change, as its a little confusing when logging into the VM that log is in /tmp. I think it might be better to separate the things done early in the VM creation (which I guess is the reason for this) into a separate log file from the actual tool logs. Maybe have a startup.log and run.log.  I think the former will be checked far less than the latter.

cluster/src/main/java/com/hartwig/pipeline/execution/vm/ComputeEngine.java
```diff
@@ -42,6 +46,7 @@
 
 public class ComputeEngine implements CloudExecutor<VirtualMachineJobDefinition> {
     private final static String APPLICATION_NAME = "vm-hosted-workload";
+    private final static int NUMBER_OF_PERSISTENT_SSDS = 4;
```
pauldwolfe: Why 4? I guess 3 would already give us the 1TB of space? 

Also think the name is wrong as these are not persistent ssds right? I think a good name would be ONE_TB_LOCAL_SSD or something like that.

cluster/src/main/java/com/hartwig/pipeline/execution/vm/ComputeEngine.java
```diff
@@ -174,36 +180,54 @@ private void addNetworkInterface(Instance instance, String projectName) {
...
+        List<AttachedDisk> disks = new ArrayList<>(asList(disk));
+        if (arguments.useScratchSsds()) {
+            attachLocalSsds(disks, projectName, zone);
+        }
+        instance.setDisks(disks);
         compute.instances().attachDisk(projectName, zone, vmName, disk);
         return sourceImage;
     }
 
+    private void attachLocalSsds(List<AttachedDisk> disks, String projectName, String zone) {
```
pauldwolfe: here we use local ssd and not scratch ssd. we should be consistent

cluster/src/main/java/com/hartwig/pipeline/execution/vm/storage/LocalSsdStorageStrategy.java
```diff
@@ -0,0 +1,20 @@
+package com.hartwig.pipeline.execution.vm.storage;
+
+import static java.lang.String.format;
+import static java.util.Arrays.asList;
+
+import java.util.List;
+
+public class LocalSsdStorageStrategy implements StorageStrategy {
+    private static final int SSD_DEVICE_COUNT = 4;
```
pauldwolfe: 4 again? any relationship to the other constant?

cluster/src/main/java/com/hartwig/pipeline/Arguments.java
nedleitch: It's actually one of the two terms Google uses, I think it might be to make it clear they're transient, like a temp directory or working directory. Surprisingly Wikipedia has an article on it:

https://en.wikipedia.org/wiki/Scratch_space

cluster/src/main/java/com/hartwig/pipeline/execution/vm/BashStartupScript.java
nedleitch: I can understand that; what if we put it in a more-standard place, like `/var/log`? That's what I'd really prefer to do, then just copy it at the end. I'm not sure why I didn't do that in fact.

You're right it's done so we have a single log for the whole run.

cluster/src/main/java/com/hartwig/pipeline/execution/vm/ComputeEngine.java
nedleitch: Yeah bad naming. Also the constant in `BashStartupScript` should be shared with this one, I just realised.

I didn't comment this because we generally don't, but there were a few counterintuitive things that came up when I was getting the local SSDs working. One of them was this: if you try to assign 3 it will error out at provisioning time. I can find the reference if you want it; from memory it was to do with the number of cores provisioned. So I had to choose between 2 and 4. If cost is a real issue and we can confirm there's no operational impact, we could drop down to 2 x 375GB.

cluster/src/main/java/com/hartwig/pipeline/execution/vm/ComputeEngine.java
nedleitch: Yeah agreed. I'm happy with either, I wish they were consistent in the GCP documentation. We can decide and I'll make them consistent.

cluster/src/main/java/com/hartwig/pipeline/execution/vm/storage/LocalSsdStorageStrategy.java
nedleitch: Yeah see the conversation up top. I realised this after I pushed.

Essentially in old-school hardware terms we're connecting four disks before we boot the machine, then we're formatting them once it's booted.

I'll pull them together.

PR #49, opened 2019-08-23
DEV-890 First cut of postprocessing in GRIDSS

nedleitch:
Doesn't work but the skeleton is in place. For the
time being we're going to _not_ include the workaround
for an old GRIDSS bug found in the the existing Pv4
scripts.


nedleitch:
Abandoning.

PR #58, opened 2019-09-13
DEV-919 Allow Dataproc zone to be specified

nedleitch:
Despite the fact we're moving away from Dataproc maybe we want to keep this for now?


pauldwolfe:
I think its good to merge this while we still have the Dataproc code around.

PR #60, opened 2019-09-17
DEV-910 Refactor compute engine code

nedleitch:
Refactor the ComputeEngine code to improve separation of concerns and
make it easier to fix problems. Also correct some stability issues.

Also make it so some jobs won't use pre-emptible VMs.  Regardless of
what the user requests, some VMs are better off without the
pre-emptible VMs because they run too long. For now GRIDSS is the
only one I'm adding this to.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/execution/vm/BucketCompletionWatcher.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/execution/vm/BucketCompletionWatcher.java
@@ -1,0 +1,49 @@
+package com.hartwig.pipeline.execution.vm;
+
+import java.time.Duration;
+import java.util.List;
+
+import com.google.cloud.storage.Blob;
+import com.hartwig.pipeline.storage.RuntimeBucket;
+
+import net.jodah.failsafe.Failsafe;
+import net.jodah.failsafe.RetryPolicy;
+
+public class BucketCompletionWatcher {
+    enum State {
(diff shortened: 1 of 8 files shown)


cluster/src/main/java/com/hartwig/pipeline/execution/vm/BucketCompletionWatcher.java
```diff
@@ -0,0 +1,49 @@
...
+            State currentState = currentState(bucket, jobDefinition);
+            if (currentState.equals(State.STILL_WAITING)) {
+                return null;
+            } else {
+                return currentState;
+            }
+        });
+    }
+
+    State currentState(RuntimeBucket bucket, VirtualMachineJobDefinition jobDefinition) {
```
pauldwolfe: Maybe this could be split up even further? No code relationship between these two exposed methods

cluster/src/main/java/com/hartwig/pipeline/execution/vm/VirtualMachineJobDefinition.java
```diff
@@ -19,6 +19,11 @@ default String imageFamily() {
 
     ResultsDirectory namespacedResults();
 
+    @Value.Default
+    default boolean preemptibleCompatible() {
```
pauldwolfe: maybe just "preemtible" is more concise as a name?

PR #63, opened 2019-09-21
DEV-947 Eliminate hardcoded input/output strings

nedleitch:
Many of our tests were hardcoding the strings for the input and
output directories which frustrates maintenance and is error-prone.

References to these strings have been centralised into the
`CommonTestEntities` class and some convenience methods added. This
approach is dead-simple (so as to not create test code that's too
smart and which runs the risk of re-implementing the implementation)
but still effective at the goal of removing duplication and hard-
coding.

Those substages which invoke Tabix have been reparented under a
common ancestor class which eliminates their identical re-
implementations of a test method for testing Tabix invocation.

Also an abstraction class, `BucketInputOutput` has been added
to simplify assertion against input and output locations that
are pushed or pulled by `gsutil`.


cluster/src/test/java/com/hartwig/pipeline/testsupport/BucketInputOutput.java
```diff
@@ -0,0 +1,29 @@
+package com.hartwig.pipeline.testsupport;
+
+import static java.lang.String.format;
+
+public class BucketInputOutput {
```
pauldwolfe: Find this abstraction confusing. push/pull sounds like a kind of stack-like thing? but then resource doesn't quite fit :) I think perhaps calling push "output", pull "input" would be better.

nedleitch: I think you're right. I actually meant to change it before I pushed. I'll fix that up.

cluster/src/main/java/com/hartwig/pipeline/execution/vm/storage/LocalSsdStorageStrategy.java
```diff
@@ -16,10 +16,10 @@ public LocalSsdStorageStrategy(int numberOfSsdDevices) {
     @Override
     public List<String> initialise() {
         String raidDevice = "/dev/md0";
-        String mdadm = format("mdadm --create %s --level=0 --raid-devices=%d ", raidDevice, numberOfSsdDevices);
+        StringBuilder mdadm = new StringBuilder(format("mdadm --create %s --level=0 --raid-devices=%d ", raidDevice, numberOfSsdDevices));
```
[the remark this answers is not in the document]
nedleitch: I _knew_ you would ask. Intellij was yelling at me! It came up in the code analysis because it's inside a loop apparently.

The discussion in the accepted answer here is relevant I think:

https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java

PR #80, opened 2019-10-09
DEV-966 Add MANIFEST for output files

nedleitch:
A manifest file is now written before pushing the files from a completed run
into the final storage location. This file contains a checksum, size in bytes
and full path (relative to the bucket) of the file.

The smoke test now checks the contents of the output directory against a
capture of the manifest as it currently looks to avoid regressions, and also
checks the contents of the manifest in the source cloud storage against the
contents of the remote cloud storage.

Suspect we may want to move some of the functionality that is currently in the
smoke test into the implementation. There should be some good discussion on
that in the PR.


cluster/src/main/java/com/hartwig/pipeline/sbpapi/SbpRestApi.java
```diff
@@ -108,9 +98,9 @@ public void postFile(final SbpFileMetadata metaData) {
         }
     }
 
-    private void patch(final String sampleID, final String status, final String json, final String entityType) {
-        LOGGER.info("Patching {} id [{}] with status [{}]", entityType, sampleID, status);
-        Response response = api().path(entityType)
+    private void patch(final String sampleID, final String status, final String json) {
+        LOGGER.info("Patching {} id [{}] with status [{}]", SbpRestApi.RUNS, sampleID, status);
+        Response response = api().path(SbpRestApi.RUNS)
```
pauldwolfe: can get rid of the SbpRestApi in SbpRestApi.RUNS? or do you find it more readable?

cluster/src/main/java/com/hartwig/pipeline/sbpapi/SbpRestApi.java
```diff
@@ -108,9 +98,9 @@ public void postFile(final SbpFileMetadata metaData) {
         }
     }
 
-    private void patch(final String sampleID, final String status, final String json, final String entityType) {
-        LOGGER.info("Patching {} id [{}] with status [{}]", entityType, sampleID, status);
-        Response response = api().path(entityType)
+    private void patch(final String sampleID, final String status, final String json) {
```
pauldwolfe: now that this method only works on RUNs maybe we should rename

cluster/src/main/java/com/hartwig/pipeline/transfer/SbpFileTransfer.java
```diff
@@ -49,30 +52,24 @@ public void publish(SomaticRunMetadata metadata, SbpRun sbpRun, String sbpBucket
...
-                        .build();
-                allFiles.add(new SourceDestPair(source, dest));
+                allFiles.add(createPair(blob, sbpBucket));
             }
         }
 
+        String manifest = writeManifest(allFiles, metadata.runName());
+        Blob manifestBlob = sourceBucket.get(manifest);
+        if (manifestBlob == null) {
+            throw new RuntimeException("Failed to upload locally-generated manifest");
```
pauldwolfe: maybe can expand a bit on this error message. can imagine it would be a little confusing for an operator (ie what can an operator do in response? Where should they look?)

cluster/src/main/java/com/hartwig/pipeline/transfer/SbpFileTransfer.java
```diff
@@ -84,11 +81,42 @@ public void publish(SomaticRunMetadata metadata, SbpRun sbpRun, String sbpBucket
...
+                .provider("gs")
+                .bucket(sourceBucket.getName())
+                .path(blob.getName())
+                .md5(blob.getMd5())
+                .size(blob.getSize())
+                .build();
+        return new SourceDestPair(source, dest);
+    }
+
+    private String writeManifest(List<SourceDestPair> allFiles, String directory) {
```
pauldwolfe: Maybe worth extracting to another class for readability/brevity of this class.

cluster/src/test/java/com/hartwig/pipeline/smoke/SmokeTest.java
```diff
@@ -27,9 +55,72 @@ public void runFullPipelineAndCheckFinalStatus() {
...
+                .sbpS3Url("s3.us-east-2.amazonaws.com").rcloneGcpRemote(GCP_REMOTE)
+                .upload(true).build();
+
+        String destinationBucket = bucketName(arguments);
+        String setName = setName(arguments);
+
+        PipelineState state = victim.start(arguments);
         assertThat(state.status()).isEqualTo(PipelineStatus.QC_FAILED);
+
+        List<String> rcloneListing = listRemoteFilesNamesOnly(format("%s:%s/%s/", S3_REMOTE, destinationBucket, setName));
```
pauldwolfe: This is a little hairy, maybe good enough to just break out into smaller methods for readability. Or another class (ManifestAssertion?) if you have a nice idea

cluster/src/test/java/com/hartwig/pipeline/smoke/SmokeTest.java
```diff
@@ -39,4 +130,15 @@ private static String workingDir() {
     private static String noDots(final String version) {
         return version.replace(".", "");
     }
+
+    private static String setName(Arguments arguments) throws IOException {
+        List<SbpSet> sets =
+                ObjectMappers.get().readValue(SbpRestApi.newInstance(arguments).getSet(SET_ID), new TypeReference<List<SbpSet>>() {
```
pauldwolfe: probably better to pass a single instance of the rest api in to these two methods than creating two separate connections.

PR #81, opened 2019-10-15
DEV-986 Small changes for JDK 11

pauldwolfe:
Needed to upgrade the mockito dependency. Then removed the explicit
setting of openjdk8 for the travis build and bump the compiler source/
target for maven.


.travis.yml
```diff
@@ -3,8 +3,6 @@ branches:
   - master
 sudo: required
 language: java
-jdk:
-- openjdk8
```
nedleitch: Do we not want to pin to 11 here?

PR #86, opened 2019-10-28
DEV-972 Google-based archiver

nedleitch:
Use the Google Transfer API to push the results of a run to a different
bucket in GCP for archival purposes.

As part of this, add command line arguments to allow separate credentials
to be supplied for the archive project. In production there will be a
separate project to store the archived runs.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/transfer/google/RetryHttpInitializerWrapper.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/transfer/google/RetryHttpInitializerWrapper.java
@@ -1,0 +1,71 @@
+package com.hartwig.pipeline.transfer.google;
+
+import java.util.logging.Logger;
+
+import com.google.api.client.auth.oauth2.Credential;
+import com.google.api.client.http.HttpBackOffIOExceptionHandler;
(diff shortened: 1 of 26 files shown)


cluster/src/main/java/com/hartwig/pipeline/transfer/google/RetryHttpInitializerWrapper.java
```diff
@@ -0,0 +1,71 @@
...
+import com.google.api.client.http.HttpBackOffIOExceptionHandler;
+import com.google.api.client.http.HttpBackOffUnsuccessfulResponseHandler;
+import com.google.api.client.http.HttpRequest;
+import com.google.api.client.http.HttpRequestInitializer;
+import com.google.api.client.http.HttpUnsuccessfulResponseHandler;
+import com.google.api.client.util.ExponentialBackOff;
+import com.google.api.client.util.Sleeper;
+import com.google.common.base.Preconditions;
+
+/**
```
pauldwolfe: was this class copied from somewhere (javadoc?)?

nedleitch: Yeah. Grabbed straight from the documentation. I'll tidy.

PR #87, opened 2019-10-29
DEV-1008 Add SAGE hotspot files to output

nedleitch:
Also make it so environment overrides can be specified to make it
so the smoke test can be run locally without much work.

### DIFF
--- a/cluster/src/test/java/com/hartwig/pipeline/smoke/LocalOverrides.java
+++ b/cluster/src/test/java/com/hartwig/pipeline/smoke/LocalOverrides.java
@@ -1,0 +1,72 @@
+package com.hartwig.pipeline.smoke;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


cluster/src/test/java/com/hartwig/pipeline/smoke/LocalOverrides.java
```diff
@@ -0,0 +1,72 @@
...
+            LOGGER.info("No local overrides will be loaded for smoke test");
+        }
+    }
+
+    public String get(String key, String defaultValue) {
+        return overrides.getOrDefault(key, defaultValue);
+    }
+
+    @JsonDeserialize
+    private static class OverridesData {
```
pauldwolfe: i think just Overrides is a descriptive enough classname

pauldwolfe: Also you can use @Immutables here to reduce code

PR #89, opened 2019-10-29
DEV-1014 Use GRIDSS copy of Picard SortVcf

nedleitch:


cluster/src/main/java/com/hartwig/pipeline/calling/structural/gridss/command/SortVcf.java
```diff
@@ -0,0 +1,34 @@
+package com.hartwig.pipeline.calling.structural.gridss.command;
+
+import java.util.List;
+
+import com.google.common.collect.ImmutableList;
+
+public class SortVcf extends GridssCommand {
+    private String withBealn;
```
pauldwolfe: final?

PR #90, opened 2019-10-31
DEV-972 Replace transfer API-based archive

nedleitch:
There were too many administrative hurdles to using the previous
implementation so replace it with rclone. The same remote is used
(but a different bucket) so there are minimal changes and the new
code is actually very simple.

One of the reasons for using the transfer API to start with was
reluctance to trust everything to rclone, which is what we're now
doing (though the transfer API was being trusted to do everything
in that case). I'm pushing this functioning version for some feedback
as I'm not sure whether we might want to do more explicit validation
after the files are in the archive location.

Also when we were testing the transfer API implementation the config
of the development storage buckets was amended to make it more
production-like. Due to this the smoke test began to have the same
error we had seen in staging and I elected to update the code so
that the manifest is copied directly to S3 rather than jumping through
GCP first, to avoid the problem.


cluster/src/main/java/com/hartwig/pipeline/transfer/sbp/SbpFileTransfer.java
```diff
@@ -60,33 +62,56 @@ public void publish(SomaticRunMetadata metadata, SbpRun sbpRun, String sbpBucket
...
-            sbpApi.postFile(metaData);
+            doRemoteWork(pair.source.toUrl(), Integer.parseInt(sbpRun.id()), pair.source.size(), pair.source.md5(), pair.dest);
         }
 
         if (arguments.cleanup()) {
             sourceObjects.forEach(Blob::delete);
         }
     }
 
+    private void writeManifest(List<SourceDestPair> allFiles, String sbpBucket, String directory, SbpRun sbpRun) {
```
pauldwolfe: It looks like the manifest will only get written to S3 and not to GCP? Why the distinction? 

Also good to keep in mind that there will be a time when we only transfer to GCP, but still need to update the SBP API with files/md5s/etc for data sharing. So I guess quite a bit of this will change when that happens.

cluster/src/test/java/com/hartwig/pipeline/smoke/SmokeTest.java
```diff
@@ -91,7 +88,10 @@ public void runFullPipelineAndCheckFinalStatus() throws Exception {
                 .sbpS3Url("s3.us-east-1.amazonaws.com")
                 .rcloneGcpRemote(GCP_REMOTE)
                 .upload(true)
-                .cleanup(false).archiveBucket("smoke-test-archive-bucket")
+                .cleanup(true)
+                //.archiveBucket("hmf-output-test")
```
pauldwolfe: this meant to be commented out?

PR #91, opened 2019-11-02
DEV-987 Move single sample run info to top level

nedleitch:
When a run is invoked as a simple sample "standalone" run, put the run's logs
and associated metadata at the top level of the output directory, rather than
one level further down in the sample's directory.


cluster/src/main/java/com/hartwig/pipeline/PipelineMain.java
```diff
@@ -84,12 +88,13 @@ private static SomaticPipeline somaticPipeline(final Arguments arguments, final
     }
 
     private static SingleSamplePipeline singleSamplePipeline(final Arguments arguments, final GoogleCredentials credentials,
-            final Storage storage, final SingleSampleEventListener eventListener) throws Exception {
+            final Storage storage, final SingleSampleEventListener eventListener, final Boolean isStandalone) throws Exception {
```
pauldwolfe: i think isSingleSample is more descriptive for this parameter

PR #96, opened 2019-11-14
DEV-1045 Upgrade to gridss 2.7.2 and use driver script

pauldwolfe:
The upgrade is relatively straightforward, just a version bump.

Moving to the driver script is a huge simplification. We can remove all
our own implementation of the gridss pipeline and just give it inputs
and consume outputs into the somatic filtering and annotation. This
commit removes much of the custom gridss code.

Also of note we include samtools as the driver script prefers this to
sambamba for greater compatibility across CPU types.


cluster/src/main/java/com/hartwig/pipeline/calling/structural/StructuralCaller.java
```diff
@@ -83,52 +81,37 @@ public String namespace() {
     public List<BashCommand> commands(final SomaticRunMetadata metadata, final Map<String, ResourceDownload> resources) {
         List<BashCommand> commands = new ArrayList<>();
         commands.add(new ExportVariableCommand("PATH", format("${PATH}:%s", dirname(new BwaCommand().asBash()))));
+        commands.add(new ExportVariableCommand("PATH",
+                format("${PATH}:%s", dirname(new VersionedToolCommand("samtools", "samtools", Versions.SAMTOOLS).asBash()))));
```
nedleitch: Does the driver script set the path for `bwa` itself or is that no longer required somehow?

cluster/src/main/java/com/hartwig/pipeline/calling/structural/gridss/stage/Driver.java
```diff
@@ -0,0 +1,56 @@
...
+                "-o",
+                output.path(),
+                "-a",
+                assemblyBamPath,
+                "-w",
+                VmDirectories.OUTPUT,
+                "-r",
+                referenceGenomePath,
+                "-j",
+                VmDirectories.TOOLS + "/" + GRIDSS + "/" + Versions.GRIDSS + "/gridss.jar",
```
nedleitch: This looks odd to my eyes, maybe especially with the `VersionedToolCommand` that it's part of, feels like you're re-implementing that with this line. I don't know if it would read better to extract the path right up to the file extension  (`.sh` or `.jar`) and then ditch the versioned command.

cluster/src/test/java/com/hartwig/pipeline/calling/structural/StructuralCallerTest.java
```diff
@@ -68,52 +68,19 @@ protected String expectedRuntimeBucketName() {
...
+                "java -Xmx32G -Dsamjdk.create_index=true -Dsamjdk.use_async_io_read_samtools=true -Dsamjdk.use_async_io_write_samtools=true -Dsamjdk.use_async_io_write_tribble=true -Dsamjdk.buffer_size=4194304 -cp /opt/tools/gridss/2.7.2/gridss.jar picard.vcf.SortVcf I=/data/output/tumor.repeatmasker_annotation.withbealn.vcf I=/data/output/tumor.repeatmasker_annotation.withannotation.vcf O=/data/output/tumor.viral_annotation.vcf.gz",
+                "(gunzip -c /data/output/tumor.viral_annotation.vcf.gz | awk ' { if (length($0) >= 4000) { gsub(\":0.00:\", \":0.0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000:\")} ; print $0  } ' > /data/output/tumor.viral_annotation.vcf)",
+                "Rscript /opt/tools/gridss/2.7.2/gridss_somatic_filter.R -p /data/resources -i /data/output/tumor.viral_annotation.vcf -o /data/output/tumor.gridss.somatic.vcf -f /data/output/tumor.gridss.somatic.full.vcf -s /opt/tools/gridss/2.7.2",
```
nedleitch: I don't think I can express tersely how happy I am with this cleanup.

PR #100, opened 2019-11-26
DEV-1060 First stab at versioning images

pauldwolfe:
Changes the image name from "diskimager" to "pipeline5" and uses a
version instead of a "type" (was always standard anyway).

The code references the image via a hardcoded version for now, as
dynamic version resolution would cause issues with local-SNAPSHOT. This
can hopefully be improved when we tie the image and the build.


cluster/src/main/java/com/hartwig/pipeline/execution/vm/VirtualMachineJobDefinition.java
```diff
@@ -2,13 +2,14 @@
...
 import com.hartwig.pipeline.execution.JobDefinition;
+import com.hartwig.pipeline.tools.Versions;
 
 import org.immutables.value.Value;
 
 @Value.Immutable
 public interface VirtualMachineJobDefinition extends JobDefinition<VirtualMachinePerformanceProfile> {
 
-    String STANDARD_IMAGE = "diskimager-standard";
+    String STANDARD_IMAGE = "pipeline5-" +Versions.imageVersion();
```
nedleitch: Could we just move all this into `Versions`?

pauldwolfe: Good idea

PR #102, opened 2019-12-04
DEV-1074 Fix single-run metadata handling

nedleitch:
Make the location of metadata depend upon the contents of the API rather
than the command line arguments. Also remove the appropriate command-line
arguments which I understand should no longer be used in the field.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/CommandLineOptions.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/CommandLineOptions.java
@@ -49,7 +49,6 @@ public class CommandLineOptions {
     private static final String ARCHIVE_BUCKET_FLAG = "archive_bucket";
     private static final String ARCHIVE_PROJECT_FLAG = "archive_project";
     private static final String ARCHIVE_PRIVATE_KEY_FLAG = "archive_private_key_path";
-    private static final String MODE_FLAG = "mode";
     private static final String SET_ID_FLAG = "set_id";
     private static final String SBP_RUN_ID_FLAG = "sbp_run_id";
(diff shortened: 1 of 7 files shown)


cluster/src/main/java/com/hartwig/pipeline/CommandLineOptions.java
```diff
@@ -1,17 +1,12 @@
...
-import org.apache.commons.cli.Options;
-import org.apache.commons.cli.ParseException;
+import org.apache.commons.cli.*;
 import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Optional;
+
 public class CommandLineOptions {
```
pauldwolfe: I think there must be some unused methods and constants in this class now that you've removed MODE. I also thing you need to remove it from the options list.

cluster/src/main/java/com/hartwig/pipeline/metadata/SomaticRunMetadata.java
```diff
@@ -29,6 +29,11 @@ static String truncate(final String sample) {
         return sample.length() > MAX_SAMPLE_LENGTH ? sample.substring(0, MAX_SAMPLE_LENGTH) : sample;
     }
 
+    @JsonIgnore
```
pauldwolfe: I think there is also an immutable annotation for this @Default to exclude it from builders, etc.

cluster/src/main/java/com/hartwig/pipeline/CommandLineOptions.java
nedleitch: Bah. Good catch.

PR #103, opened 2019-12-11
DEV-1081 Improve batch5 usability and extensibility

nedleitch:


batch/src/main/java/com/hartwig/batch/BatchArguments.java
```diff
@@ -20,10 +25,16 @@
...
 
+    String INPUT_FORMAT = "input_format";
+    String JSON_INPUT = "json";
+    String FLAT_INPUT = "flat";
+
     int concurrency();
 
     String inputFile();
 
+    String inputFormat();
```
pauldwolfe: maybe we should use an enum here?

batch/src/main/java/com/hartwig/batch/BatchDispatcher.java
```diff
@@ -32,48 +32,52 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(BatchDispatcher.class);
     private final BatchArguments arguments;
     private final InstanceFactory instanceFactory;
+    private InputParser inputParser;
```
pauldwolfe: fields should be final

batch/src/main/java/com/hartwig/batch/input/FlatInputParser.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.commons.io.FileUtils;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class FlatInputParser implements InputParser {
+    private String inputFile;
```
pauldwolfe: both implementations take inputfile and billedproject as fields, perhaps best to move these into arguments to parse.

batch/src/main/java/com/hartwig/batch/operations/SagePon.java
```diff
@@ -72,12 +74,12 @@ public VirtualMachineJobDefinition execute(final InputFileDescriptor descriptor,
...
 
         // Download bams
-        startupScript.addCommand(() -> "touch /data/input/files.list");
-        startupScript.addCommand(() -> format("echo %s | tee -a  /data/input/files.list", gcTumorBamFile));
-        startupScript.addCommand(() -> format("echo %s.bai | tee -a /data/input/files.list", gcTumorBamFile));
-        startupScript.addCommand(() -> format("echo %s | tee -a /data/input/files.list", gcReferenceBamFile));
-        startupScript.addCommand(() -> format("echo %s.bai | tee -a /data/input/files.list", gcReferenceBamFile));
-        startupScript.addCommand(() -> format("cat /data/input/files.list | gsutil -m -u hmf-crunch cp -I %s", "/data/input/"));
+        startupScript.addCommand(() -> "touch /data/inputs/files.list");
+        startupScript.addCommand(() -> format("echo %s | tee -a  /data/inputs/files.list", gcTumorBamFile));
```
pauldwolfe: could remove this duplication by adding an Echo/Tee class command.

batch/src/main/java/com/hartwig/batch/input/FlatInputParser.java
nedleitch: Good catch. Done.

PR #112, opened 2020-01-09
DEV-1101 Integrate pipeline5 with bcl2fastq on GCP

pauldwolfe:
Adds a toggle that once enabled, will upload fastq input from GCP rather than the SBP S3.
The location of the fastq is still taken from the SBP API (run->sample->fastq).

Also updated the Bcl2Fastq command thread settings to match production.

### DIFF
--- a/bcl2fastq/src/main/java/com/hartwig/bcl2fastq/conversion/ResultAggregation.java
+++ b/bcl2fastq/src/main/java/com/hartwig/bcl2fastq/conversion/ResultAggregation.java
@@ -14,38 +14,38 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
 
-import com.google.api.gax.paging.Page;
 import com.google.cloud.storage.Blob;
-import com.google.cloud.storage.Bucket;
-import com.google.cloud.storage.Storage;
 import com.hartwig.bcl2fastq.samplesheet.IlluminaSample;
 import com.hartwig.bcl2fastq.samplesheet.SampleSheet;
 import com.hartwig.bcl2fastq.stats.LaneStats;
 import com.hartwig.bcl2fastq.stats.SampleStats;
 import com.hartwig.bcl2fastq.stats.Stats;
 import com.hartwig.bcl2fastq.stats.UndeterminedStats;
+import com.hartwig.pipeline.ResultsDirectory;
+import com.hartwig.pipeline.storage.RuntimeBucket;
 
 public class ResultAggregation {
 
-    private final Bucket bucket;
+    private final RuntimeBucket bucket;
+    private final ResultsDirectory resultsDirectory;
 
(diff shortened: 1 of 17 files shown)


bcl2fastq/src/main/java/com/hartwig/bcl2fastq/metadata/SbpFastqMetadataApi.java
```diff
@@ -55,7 +55,7 @@ public SbpSample findOrCreate(String barcode, String submission) {
                 try {
                     Response response = samples().request()
                             .post(Entity.entity(objectMapper.writeValueAsString(sample), MediaType.APPLICATION_JSON_TYPE));
-                    if (response.getStatus() == 200) {
+                    if (response.getStatus() == 201) {
```
nedleitch: `Response.Status.CREATED` or something similar?

cluster/src/main/java/com/hartwig/pipeline/execution/vm/VirtualMachinePerformanceProfile.java
```diff
@@ -1,24 +1,26 @@
...
-public interface VirtualMachinePerformanceProfile extends PerformanceProfile{
+public interface VirtualMachinePerformanceProfile extends PerformanceProfile {
+
+    @Value.Default
+    default String uri() {
+        return machineType().uri();
+    }
 
-    String uri();
+    MachineType machineType();
```
nedleitch: Much better.

bcl2fastq/src/main/java/com/hartwig/bcl2fastq/conversion/ResultAggregation.java
```diff
@@ -95,7 +95,15 @@ static ConvertedFastq fastq(IlluminaSample sample, Map.Entry<String, List<Blob>>
     }
 
     private static String outputPath(final IlluminaSample sample, final Stats stats, final Blob blobR1) {
-        return stats.flowcell() + "/" + sample.barcode() + "/" + new File(blobR1.getName()).getName();
+        String[] fileNameSplit = new File(blobR1.getName()).getName().split("_");
```
nedleitch: We have had surprises in the past, would it be wise to check the filename conforms to expectations?

bcl2fastq/src/main/java/com/hartwig/bcl2fastq/metadata/SbpFastqMetadataApi.java
pauldwolfe: Good suggestion

bcl2fastq/src/main/java/com/hartwig/bcl2fastq/conversion/ResultAggregation.java
pauldwolfe: hmm, as this file convention is produced by the version of bcl2fastq also in this codebase I think its fine to have this implicit dependency on the filename.

PR #113, opened 2020-01-13
DEV-1106 Always use a second PD SSD when local SSDs disabled

pauldwolfe:
This solves the issue of the 2TB limit of the boot disk. The boot disk will
always be 100GB (although perhaps that can be reduced?) and workind disk size
the size of the new disk.

The disk is mounted in the startup script with the PersistentStorageStrategy.
this mounts /dev/sdb to /data.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/execution/vm/ComputeEngine.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/execution/vm/ComputeEngine.java
@@ -32,6 +32,7 @@ import com.google.auth.oauth2.GoogleCredentials;
 import com.hartwig.pipeline.CommonArguments;
 import com.hartwig.pipeline.execution.PipelineStatus;
 import com.hartwig.pipeline.execution.vm.storage.LocalSsdStorageStrategy;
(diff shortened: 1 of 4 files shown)


cluster/src/main/java/com/hartwig/pipeline/execution/vm/ComputeEngine.java
```diff
@@ -203,33 +205,46 @@ private void addNetworkInterface(Instance instance, String projectName) {
...
+    private Image attachDisks(Compute compute, Instance instance, VirtualMachineJobDefinition jobDefinition, String projectName,
+            String vmName, String zone) throws IOException {
         Image sourceImage = resolveLatestImage(compute, jobDefinition.imageFamily(), projectName);
-        AttachedDisk disk = new AttachedDisk();
-        disk.setBoot(true);
-        disk.setAutoDelete(true);
+        AttachedDisk bootDisk = new AttachedDisk();
+        bootDisk.setBoot(true);
+        bootDisk.setAutoDelete(true);
         AttachedDiskInitializeParams params = new AttachedDiskInitializeParams();
```
nedleitch: Nit: maybe extend the analogy and call this `bootDiskParams`?

PR #130, opened 2020-02-21
DEV-1122 Add CRAM conversion to pipeline

nedleitch:


cluster/src/main/java/com/hartwig/pipeline/cram/CramAndValidateCommands.java
```diff
@@ -0,0 +1,44 @@
...
+import com.hartwig.pipeline.execution.vm.Bash;
+import com.hartwig.pipeline.execution.vm.BashCommand;
+import com.hartwig.pipeline.execution.vm.JavaClassCommand;
+import com.hartwig.pipeline.resource.Resource;
+import com.hartwig.pipeline.tools.Versions;
+
+import java.util.Collections;
+import java.util.List;
+
+public class CramAndValidateCommands {
```
pauldwolfe: I think the Substage interface would work well for this

cluster/src/main/java/com/hartwig/pipeline/cram/CramConversion.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public CramOutput output(SingleSampleRunMetadata metadata, PipelineStatus jobStatus, RuntimeBucket bucket, ResultsDirectory resultsDirectory) {
+        String cram = CramOutput.cramFile(metadata.sampleName());
+        String crai = CramOutput.craiFile(metadata.sampleName());
+        Folder folder = Folder.from(metadata);
+        return CramOutput.builder()
+                .status(jobStatus)
+                .addReportComponents(
```
pauldwolfe: I see we add the CRAM here to the final output I guess we want to remove the BAM from the aligner report components as well?

pauldwolfe: Or maybe you were thinking we persist both to start?

cluster/src/main/java/com/hartwig/pipeline/cram/CramConversion.java
```diff
@@ -0,0 +1,85 @@
...
+                .build();
+    }
+
+    @Override
+    public CramOutput skippedOutput(SingleSampleRunMetadata metadata) {
+        throw new IllegalStateException("CRAM conversion cannot be skipped.");
+    }
+
+    @Override
+    public boolean shouldRun(Arguments arguments) {
```
pauldwolfe: I think we should have an option to enable/disable this feature

cluster/src/main/java/com/hartwig/pipeline/cram/CramConversion.java
nedleitch: I was. But maybe that's overly cautious and not really beneficial.

batch/src/main/java/com/hartwig/batch/operations/SamtoolsCramToBam.java
```diff
@@ -1,53 +1,41 @@
...
+import com.hartwig.pipeline.execution.vm.unix.MvCommand;
 import com.hartwig.pipeline.storage.GoogleStorageLocation;
 import com.hartwig.pipeline.storage.RuntimeBucket;
-import com.hartwig.pipeline.tools.Versions;
+
+import java.io.File;
+
+import static java.lang.String.format;
 
 public class SamtoolsCramToBam implements BatchOperation {
```
pauldwolfe: Rename to SamtoolsBamToCram?

cluster/src/main/java/com/hartwig/pipeline/cram/CramAndValidateCommands.java
pauldwolfe: Still do :)

cluster/src/main/java/com/hartwig/pipeline/tools/Versions.java
```diff
@@ -30,6 +30,7 @@
     String CHORD = "60.02_1.03";
     String SAMTOOLS = "1.9";
     String BCL2FASTQ = "2.20.0.422";
+    String BAMCOMP = "0.1";
```
pauldwolfe: 0.1 implies its not ready for production, would suggest we use 1.0 once its validated (which I think it sort of is by Korneel)

batch/src/main/java/com/hartwig/batch/operations/SamtoolsCramToBam.java
nedleitch: Good call. Not sure why this was done this way.

cluster/src/main/java/com/hartwig/pipeline/cram/CramAndValidateCommands.java
nedleitch: Sorry wasn't ignoring this suggestion, just forgot. But now that I look more closely at it, I'd have to do some gymnastics to use the substages from the batch side because it doesn't really care about samples (it deals generically with input files) and I'm not sure whether it's worth the effort. For the pipeline invocation it makes more sense but not sure whether that's compelling by itself.

pauldwolfe: I don't think it would be too hard to remove the references to samples, but also fine to leave this for now and make that a part of another separate refactoring task.

PR #135, opened 2020-03-03
DEV-1123 Make polling interval configurable

nedleitch:
Batch jobs with large numbers of VMs running can overwhelm our global
API request limits. With this change we can drop the resolution of
state change discovery in favour of the ability to run more concurrent
VMs.


cluster/src/main/java/com/hartwig/pipeline/Arguments.java
```diff
@@ -17,120 +17,7 @@
 
     boolean cleanup();
 
-    static ImmutableArguments.Builder defaultsBuilder(String profileString) {
```
pauldwolfe: Why is all this marked as deleted then added again?

PR #138, opened 2020-03-09
DEV-1163 Make CMEK a required argument

nedleitch:
Better to require every invocation to have a key (even if it is just a
development one) than risk not having one on a run that requires it.


batch/src/main/java/com/hartwig/batch/BatchArguments.java
```diff
@@ -41,8 +33,8 @@ static BatchArguments from(String[] args) {
             CommandLine commandLine = new DefaultParser().parse(options(), args);
             return ImmutableBatchArguments.builder()
                     .command(args[0])
-                    .project(commandLine.getOptionValue(PROJECT, "hmf-pipeline-development"))
-                    .region(commandLine.getOptionValue(REGION, "europe-west4"))
+                    .project(commandLine.getOptionValue(PROJECT, Arguments.DEFAULT_DEVELOPMENT_PROJECT))
```
pauldwolfe: better to pull these up into CommonArguments than to introduce a dependency on pipeline Arguments

cluster/src/main/java/com/hartwig/pipeline/Arguments.java
```diff
@@ -17,6 +17,120 @@
 
     boolean cleanup();
 
+    String DEFAULT_DOCKER_CMEK = "/secrets/storage-cmek.json";
```
pauldwolfe: Normal the CMEK is actually just the name of the key used ie. projects/project/location/blah/blah. Does this actually work having it a json file? I don't see why running in Docker would need to impact CMEK (it's more like the prokect or location setttings).

cluster/src/main/java/com/hartwig/pipeline/Arguments.java
```diff
@@ -17,6 +17,120 @@
...
+
+    boolean upload();
```
pauldwolfe: Why are these all "new lines"?

cluster/src/main/java/com/hartwig/pipeline/storage/RuntimeBucket.java
```diff
@@ -49,10 +48,11 @@ private synchronized static RuntimeBucket createBucketIfNeeded(final Storage sto
             LOGGER.debug("Creating runtime bucket [{}] in Google Storage", runId);
             BucketInfo.Builder builder =
                     BucketInfo.newBuilder(runId).setStorageClass(StorageClass.REGIONAL).setLocation(arguments.region());
-            arguments.cmek().ifPresent(key -> {
-                LOGGER.info("Using CMEK key [{}] to encrypt all buckets", key);
-                builder.setDefaultKmsKeyName(key);
-            });
+            if ("".equals(arguments.cmek().trim())) {
```
pauldwolfe: use cmek().trim().isEmpty()

cluster/src/main/java/com/hartwig/pipeline/CommonArguments.java
```diff
@@ -39,7 +39,7 @@
 
     String serviceAccountEmail();
 
-    Optional<String> cmek();
```
pauldwolfe: As you pointed out this doesn't make it very external friendly. But... I don't think that's the most important thing right now so we can decide later whether its worth the additional complexity of having a separate flag to disable cmek and make this an optional again.

cluster/src/main/java/com/hartwig/pipeline/storage/RuntimeBucket.java
nedleitch: Yeah that's weak.

cluster/src/main/java/com/hartwig/pipeline/Arguments.java
nedleitch: Ah. I moved the builder method below the constants and interface methods.

PR #140, opened 2020-03-19
DEV-1153 Add HG 38 Support

pauldwolfe:
I just pushed my branch back to master - DEV-1153 - so feel free to look at the changes. They all revolve around:
- optionally providing a ref genome version argument to the pipeline, with default of HG37
- adding in a layer in many of the resource sub-directories to store the files for each version
- adding an interface to get resource files which handles this versioning under the hood (not bonnet)


cluster/src/main/java/com/hartwig/pipeline/calling/structural/StructuralCaller.java
```diff
@@ -77,18 +81,27 @@ public String namespace() {
         commands.add(new ExportVariableCommand("PATH",
                 format("${PATH}:%s", dirname(new VersionedToolCommand("samtools", "samtools", Versions.SAMTOOLS).asBash()))));
 
+        // TEMP
```
pauldwolfe: This necessary still? The resources in the image are static we should just be able to set this up exactly how we want it and remove the need for checking files and copying.

cluster/src/main/java/com/hartwig/pipeline/resource/Hg37ResourceFiles.java
```diff
@@ -0,0 +1,52 @@
...
+    public String versionDirectory() { return HG37_DIRECTORY; }
+
+    private String formPath(String name, String file) {
+        return String.format("%s/%s/%s/%s", VmDirectories.RESOURCES, name, versionDirectory(), file);
+    }
+
+    private static final String REF_GENOME_FASTA_HG37_FILE = "Homo_sapiens.GRCh37.GATK.illumina.fasta";
+
+    public static final String REFERENCE_GENOME_FASTA_HG37 =
+            String.format("%s/%s/%s/%s", VmDirectories.RESOURCES, REFERENCE_GENOME, HG37_DIRECTORY, REF_GENOME_FASTA_HG37_FILE);
```
jonbaber: Not sure why there is public static method. I think it makes more sense for the batch applications to access this via the ResourcesFilesFactory same as the normal pipeline.

cluster/src/main/java/com/hartwig/pipeline/resource/Hg37ResourceFiles.java
```diff
@@ -0,0 +1,52 @@
...
+import static com.hartwig.pipeline.resource.ResourceNames.GRIDSS_REPEAT_MASKER_DB;
+import static com.hartwig.pipeline.resource.ResourceNames.MAPPABILITY;
+import static com.hartwig.pipeline.resource.ResourceNames.REFERENCE_GENOME;
+import static com.hartwig.pipeline.resource.ResourceNames.SAGE;
+import static com.hartwig.pipeline.resource.ResourceNames.SNPEFF;
+
+import com.hartwig.pipeline.execution.vm.VmDirectories;
+
+public class Hg37ResourceFiles implements ResourceFiles
+{
```
jonbaber: Please apply the code style of the project. The project root directory contains hmf_code_schema.xml if you want to have it done automatically for this project.

cluster/src/main/java/com/hartwig/pipeline/resource/Hg37ResourceFiles.java
charlesshale: Removed this method

cluster/src/main/java/com/hartwig/pipeline/resource/Hg37ResourceFiles.java
charlesshale: Changed

PR #142, opened 2020-04-08
DEV-1194: Use SAGE as somatic caller in HG38

jonbaber:


cluster/src/main/java/com/hartwig/pipeline/CommandLineOptions.java
```diff
@@ -290,7 +292,7 @@ public static Arguments from(String[] args) throws ParseException {
                     .runAligner(booleanOptionWithDefault(commandLine, RUN_ALIGNER_FLAG, defaults.runAligner()))
                     .runSnpGenotyper(booleanOptionWithDefault(commandLine, RUN_SNP_GENOTYPER_FLAG, defaults.runSnpGenotyper()))
                     .runGermlineCaller(booleanOptionWithDefault(commandLine, RUN_GERMLINE_CALLER_FLAG, defaults.runGermlineCaller()))
-                    .runSomaticCaller(booleanOptionWithDefault(commandLine, RUN_SOMATIC_CALLER_FLAG, defaults.runSomaticCaller()))
+                    .runSomaticCaller(refGenomeVersion.equals(RefGenomeVersion.HG37) && booleanOptionWithDefault(commandLine, RUN_SOMATIC_CALLER_FLAG, defaults.runSomaticCaller()))
```
pauldwolfe, line 295: Thinking a better place for this would be in the shouldRun method of the somatic caller. Just that we don't have any other logic in this class and its also hard to unit test it.

jonbaber: Good idea.

PR #143, opened 2020-04-24
DEV-1245 Create a JSON sample source

pauldwolfe:
Create a new class JsonSampleSource which parses a JSON representation of the sample/lane/fastq model, passing the GCS locations along as paths within the lanes. This combined with the GS upload allows us to put fastq files in arbitrary locations.

Eliminated the file system upload as its no longer used (local runs should use this new way, or the acc api).

Also eliminated the patient module and some interfaces which didn't really add any value. Included in this commit as it made the addition of the JsonSampleSource more straightforward.


cluster/src/main/java/com/hartwig/pipeline/CommandLineOptions.java
```diff
@@ -64,7 +63,8 @@
     private static final String CMEK_FLAG = "cmek";
     private static final String SHALLOW_FLAG = "shallow";
     private static final String ZONE_FLAG = "zone";
-    private static final String REF_GENOME_VERSION = "ref_genome_version";
+    private static final String REF_GENOME_VERSION_FLAG = "ref_genome_version";
+    private static final String JSON_FLAG = "json";
```
nedleitch, line 67: Could we come up with a name that is more descriptive of its function rather than just indicative of its file format? Like maybe `sample_json` or something?

cluster/src/main/java/com/hartwig/patient/Sample.java
```diff
@@ -0,0 +1,35 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonDeserialize(as = ImmutableSample.class)
+@Value.Style(jdkOnly = true)
+public interface Sample {
+
+    String NOT_APPLICABLE = "NA";
```
nedleitch, line 14: We're going to need to make a "not applicable" constant somewhere central...

cluster/src/main/java/com/hartwig/pipeline/CommandLineOptions.java
pauldwolfe: sample_json sounds reasonable. I'll use that.

PR #149, opened 2020-05-01
DEV-1250 Register CRAMs with API

nedleitch:
Allow calling of arbitrary API endpoints after files have been created
so that we can add additional metadata for certain files from the run.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/cram/CramConversion.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/cram/CramConversion.java
@@ -18,7 +18,6 @@ import com.hartwig.pipeline.execution.vm.VirtualMachineJobDefinition;
 import com.hartwig.pipeline.execution.vm.VirtualMachinePerformanceProfile;
 import com.hartwig.pipeline.execution.vm.VmDirectories;
 import com.hartwig.pipeline.metadata.AddDatatypeToFile;
-import com.hartwig.pipeline.metadata.AdditionalApiCalls;
 import com.hartwig.pipeline.metadata.LinkFileToSample;
 import com.hartwig.pipeline.metadata.SingleSampleRunMetadata;
 import com.hartwig.pipeline.report.Folder;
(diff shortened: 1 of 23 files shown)


cluster/src/main/java/com/hartwig/pipeline/cram/CramConversion.java
```diff
@@ -66,10 +71,18 @@ public CramOutput output(SingleSampleRunMetadata metadata, PipelineStatus jobSta
         String cram = new File(outputCram).getName();
         String crai = CramOutput.craiFile(cram);
         Folder folder = Folder.from(metadata);
+
+        String fullCram = format("%s%s/%s", folder.name(), NAMESPACE, cram);
+        String fullCrai = format("%s%s/%s", folder.name(), NAMESPACE, crai);
+
+        AdditionalApiCalls.instance().register(fullCram, new LinkFileToSample(metadata.sampleId()));
```
pauldwolfe, line 78: singleton anti-pattern. makes the class impossible to unit test. let's invert the dependency and make AdditionalApiCalls a field.

Also I was thinking we'd use the "ReportComponents" for this rather than introducing a new class in which to register them.

cluster/src/main/java/com/hartwig/pipeline/metadata/AdditionalApiCalls.java
```diff
@@ -0,0 +1,47 @@
...
+    private AdditionalApiCalls() {
+        metadata = new HashMap<>();
+        //String name = RunTag.apply(arguments, metadata.sampleId());
+    }
+
+    public static AdditionalApiCalls instance() {
+        return INSTANCE;
+    }
+
+    public synchronized void register(String filePath, ApiFileOperation extraOperation) {
```
pauldwolfe, line 28: another side-effect of singleton. Why isn't register modelled as an Operation somehow?

cluster/src/main/java/com/hartwig/pipeline/metadata/ApiFileOperation.java
```diff
@@ -0,0 +1,8 @@
+package com.hartwig.pipeline.metadata;
+
+import com.hartwig.pipeline.sbpapi.FileResponse;
+import com.hartwig.pipeline.sbpapi.SbpRestApi;
+
+public interface ApiFileOperation {
```
pauldwolfe, line 6: I like the thought behind this but feel like it introduces inconsistency to only have ApiFile operation and no similar abstractions for other operations.

cluster/src/main/java/com/hartwig/pipeline/metadata/LinkFileToSample.java
```diff
@@ -0,0 +1,17 @@
...
+
+public class LinkFileToSample implements ApiFileOperation {
+    private final String sampleId;
+
+    public LinkFileToSample(String sampleId) {
+        this.sampleId = sampleId;
+    }
+
+    @Override
+    public void apply(final SbpRestApi api, final FileResponse fileResponse) {
```
pauldwolfe, line 14: They also don't do much... maybe you can get away with just a BiFunction where you need the abstraction?

cluster/src/main/java/com/hartwig/pipeline/sbpapi/FileResponse.java
```diff
@@ -0,0 +1,8 @@
+package com.hartwig.pipeline.sbpapi;
+
+public class FileResponse {
+    public int id;
+
+    public FileResponse() {
```
pauldwolfe, line 6: empty constructor and non-final mutable field? looks like a job for immutables

cluster/src/main/java/com/hartwig/pipeline/cram/CramConversion.java
nedleitch: Meant to mention - this is absolutely just a waypoint, it won't be staying that way, working on getting rid of it now.

I was sitting on the fence with putting it in the `ReportComponents`, wasn't sure about overloading but am happy to make that change.

cluster/src/main/java/com/hartwig/pipeline/sbpapi/FileResponse.java
nedleitch: Indeed. Was actually looking for a way to inline it but will switch to an Immutable.

PR #153, opened 2020-05-20
DEV-1279 Copy all single-pipeline files on rerun

nedleitch:
We encountered a failure in production in which a failure in the CRAM
stage resulted in a rerun of the pipeline, but after the rerun succeeded
the output directory did not contain the files from the sample that had
failed the first time (none of them; the subdirectory for the sample
itself wasn't even present).

Looks like the reason for this was that the flag file which tells the
pipeline that the staging directory has been populated was left-over
from the first run, which meant that upon re-run the pipeline didn't
wait for the files to be copied from the staging bucket into the final
archive bucket.

Also improve logging output for post-applied operations to make
debugging this problem easier.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/metadata/AddDatatypeToFile.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/metadata/AddDatatypeToFile.java
@@ -1,5 +1,7 @@
 package com.hartwig.pipeline.metadata;
 
+import static java.lang.String.format;
+
 import com.hartwig.pipeline.sbpapi.AddFileApiResponse;
 import com.hartwig.pipeline.sbpapi.SbpRestApi;
 
@@ -20,5 +22,10 @@ public class AddDatatypeToFile implements ApiFileOperation {
     @Override
     public String path() {
         return path;
+    }
+
+    @Override
+    public String toString() {
+        return format("add datatype [%s] to [%s]", datatype, path);
     }
 }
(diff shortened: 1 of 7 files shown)


cluster/src/main/java/com/hartwig/pipeline/metadata/AddDatatypeToFile.java
```diff
@@ -21,4 +23,9 @@ public void apply(final SbpRestApi api, final AddFileApiResponse file) {
     public String path() {
         return path;
     }
+
+    @Override
+    public String toString() {
```
pauldwolfe, line 28: How do these toStrings relate to the bug?

cluster/src/main/java/com/hartwig/pipeline/report/PipelineResults.java
```diff
@@ -97,4 +97,12 @@ private void writeComplete(final String name) {
     private String path(final String name, final Folder folder, final String fileName) {
         return String.format("%s/%s%s", name, folder.name(), fileName);
     }
+
+    public void initialise(final Arguments arguments, final SingleSampleRunMetadata metadata) {
```
pauldwolfe, line 101: Why is the STAGING_COMPETE flag written if the single sample pipeline failed? Is there any way we can fix this such that staging complete isn't written when CRAM fails and the somatic pipeline is never kicked off?

cluster/src/main/java/com/hartwig/pipeline/metadata/AddDatatypeToFile.java
nedleitch: I mentioned in the commit message but the detail is that as part of testing this change I made the pipeline temporarily fail on only one of the CRAM operations. It was useful to be able to see that the correct one was being registered at the end of the run (all the operations that are applied already get logged) and having a proper `toString` on the operations provided that.

PR #155, opened 2020-05-25
DEV-1287 Disable sage caller when shallow run

pauldwolfe:
Currently the output is not used in the shallow run and it doesn't provide anything interesting to use later. Disable the caller when the shallow flag is true.


cluster/src/main/java/com/hartwig/pipeline/calling/somatic/SageV2Caller.java
```diff
@@ -123,6 +122,6 @@ public SomaticCallerOutput skippedOutput(final SomaticRunMetadata metadata) {
 
     @Override
     public boolean shouldRun(final Arguments arguments) {
-        return arguments.runSageCaller();
+        return !arguments.shallow() && arguments.runSageCaller();
```
nedleitch, line 125: There's no test for this condition anywhere?

PR #162, opened 2020-06-10
Dev 1201

jonbaber:


batch/src/main/java/com/hartwig/batch/operations/SageGermline.java
```diff
@@ -2,132 +2,110 @@
...
-        startupScript.addCommand(() -> "touch /data/input/files.list");
-        startupScript.addCommand(() -> format("echo %s | tee -a  /data/input/files.list", gcTumorBamFile));
-        startupScript.addCommand(() -> format("echo %s.bai | tee -a /data/input/files.list", gcTumorBamFile));
-        startupScript.addCommand(() -> format("echo %s | tee -a /data/input/files.list", gcReferenceBamFile));
-        startupScript.addCommand(() -> format("echo %s.bai | tee -a /data/input/files.list", gcReferenceBamFile));
-        startupScript.addCommand(() -> format("cat /data/input/files.list | gsutil -m -u hmf-crunch cp -I %s", "/data/input/"));
+        // Download latest jar file
+        startupScript.addCommand(() -> format("gsutil -u hmf-crunch cp %s %s",
+                "gs://batch-sage-validation/resources/sage.jar",
+                "/opt/tools/sage/" + Versions.SAGE + "/sage.jar"));
```
nedleitch, line 61: If the JAR version is stable at this point it's preferable to make an image containing it. Appreciate that during development this would be more expedient but if this code is going to stick around I'd rather it's more "official".

batch/src/main/java/com/hartwig/batch/operations/SageGermline.java
```diff
@@ -2,132 +2,110 @@
...
+        final SageV2Application sageV2Application = new SageV2Application(sageCommandBuilder);
 
-        //        startupScript.addCommand(() -> format("gsutil -u hmf-crunch cp %s %s", gcTumorBamFile, localTumorBamFile + ".bai"));
-        //        startupScript.addCommand(() -> format("gsutil -u hmf-crunch cp %s %s", gcTumorBamFile + ".bai", localTumorBamFile + ".bai"));
-        //        startupScript.addCommand(() -> format("gsutil -u hmf-crunch cp %s %s", gcReferenceBamFile, localReferenceBamFile));
-        //        startupScript.addCommand(() -> format("gsutil -u hmf-crunch cp %s %s", gcReferenceBamFile + ".bai", localReferenceBamFile + ".bai"));
+        // Convert to bam if necessary
+        if (!localReferenceFile.equals(localReferenceBam)) {
+            startupScript.addCommands(cramToBam(localReferenceFile));
+        }
```
nedleitch, line 74: You might consider re-using the existing batch commands that we used for the batch conversion, to help simplify this batch. With that approach you'd work out which BAMs need to be converted up-front, run a batch, then remove the conditional logic here.

The existing code also does some validation. See `SamtoolsBamToCram` if interested.

PR #167, opened 2020-07-02
DEV-1345 Upgrade GRIDSS to 2.9.3

pauldwolfe:
Does the normal version bump but also switches to using AnnotatedInsertedSequences to do the repeat masking annotation,
instead of the old R script.


cluster/src/main/java/com/hartwig/pipeline/resource/Hg38ResourceFiles.java
```diff
@@ -22,7 +22,7 @@ private String formPath(String name, String file) {
         return String.format("%s/%s/%s/%s", VmDirectories.RESOURCES, name, versionDirectory(), file);
     }
 
-    public String refGenomeFile() { return formPath(REFERENCE_GENOME, "Homo_sapiens_assembly38.fasta"); }
+    public String refGenomeFile() { return formPath(REFERENCE_GENOME, "GCA_000001405.15_GRCh38_no_alt_analysis_set.fna"); }
```
jonbaber, line 25: This looks a little odd? Should it be a fna file?

pauldwolfe: This is the correct actually. Peter discovered a few months ago we'd been using the wrong ref genome. This is the "no_alt_analysis" set, the prior is the full set. Thought this impacted your GC profile as well?

FNA is fasta and means fasta nucleic acid.

PR #168, opened 2020-07-18
DEV-1386 Link to datatype/sample once per file

nedleitch:
Not precisely sure yet why sometimes multiple copies of the
same operation are requested but it is not incorrect to store
and apply the operations in a set rather than a list. 

This will resolve the problem we are currently seeing 
anyway.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/PipelineState.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/PipelineState.java
@@ -1,10 +1,10 @@
 package com.hartwig.pipeline;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
 
-import com.google.common.collect.Lists;
 import com.hartwig.pipeline.execution.PipelineStatus;
 
 import org.slf4j.Logger;
@@ -13,7 +13,15 @@ import org.slf4j.LoggerFactory;
 public class PipelineState {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(PipelineState.class);
-    private final List<StageOutput> stageOutputs = Lists.newArrayList();
+    private final List<StageOutput> stageOutputs;
+
+    public PipelineState() {
+        this.stageOutputs = new ArrayList<>();
+    }
+
+    public PipelineState(final List<StageOutput> stageOutputs) {
+        this.stageOutputs = stageOutputs;
+    }
 
     public <T extends StageOutput> T add(final T stageOutput) {
         stageOutputs.add(stageOutput);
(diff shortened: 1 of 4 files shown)


cluster/src/main/java/com/hartwig/pipeline/transfer/SbpFileApiUpdate.java
```diff
@@ -30,14 +30,15 @@
     private final SbpRun sbpRun;
     private final Bucket sourceBucket;
     private final SbpRestApi sbpApi;
-    private final List<ApiFileOperation> fileOperations = new ArrayList<>();
+    private final Set<ApiFileOperation> fileOperations;
```
pauldwolfe, line 33: As neither implementation of ApiFileOperation implement equals/hashcode this is only ensuring that the same reference isn't in the collection twice, which feels like masking another bug. I think we should keep going until we find where that is happening.

Still fine to keep idea but maybe then better we check for duplicates and log something before removing them?

cluster/src/main/java/com/hartwig/pipeline/transfer/SbpFileApiUpdate.java
```diff
@@ -59,14 +62,26 @@ public void accept(final Blob blob) {
...
-                        LOGGER.info("Applying: {}", fileOperation);
+                        LOGGER.info("Applying [{}] for [{}]", fileOperation, blob.getName());
                         fileOperation.apply(sbpApi, fileResponse);
                     }
                 }
             }
         }
     }
 
+    private void storeOperations(List<StageOutput> stageOutputs) {
```
pauldwolfe, line 73: I would suggest reverting all this to get back to a place where you can duplicate the bug. I don't see a possible concurrency issue as the new code only processes the list which should already contain the duplicate. If you can't duplicate it after that, it suggests we still don't quite understand the problem.

cluster/src/main/java/com/hartwig/pipeline/PipelineState.java
```diff
@@ -13,7 +13,15 @@
...
 
     private static final Logger LOGGER = LoggerFactory.getLogger(PipelineState.class);
-    private final List<StageOutput> stageOutputs = Lists.newArrayList();
+    private final List<StageOutput> stageOutputs;
+
+    public PipelineState() {
+        this.stageOutputs = new ArrayList<>();
+    }
+
+    public PipelineState(final List<StageOutput> stageOutputs) {
```
nedleitch, line 22: [Package?] private?

cluster/src/main/java/com/hartwig/pipeline/transfer/SbpFileApiUpdate.java
pauldwolfe: I don't think we should keep this code, its a little confusing when you don't have the full context of why its here. If we do keep, the second info statement should probably be a WARN.

PR #177, opened 2020-07-30
DEV-1420 Apply CMEK to patient report bucket

nedleitch:
If the pipeline creates this bucket (as it is likely to do when we are
running Platinum) it will now check that a CMEK is specified and apply
it to the new bucket.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/report/PipelineResultsProvider.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/report/PipelineResultsProvider.java
@@ -25,10 +25,14 @@ public class PipelineResultsProvider {
     public PipelineResults get() {
         Bucket reportBucket = storage.get(arguments.patientReportBucket());
         if (reportBucket == null) {
-            reportBucket = storage.create(BucketInfo.newBuilder(arguments.patientReportBucket())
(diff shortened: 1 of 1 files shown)


cluster/src/main/java/com/hartwig/pipeline/report/PipelineResultsProvider.java
```diff
@@ -25,10 +25,14 @@ public static PipelineResultsProvider from(final Storage storage, final Argument
     public PipelineResults get() {
         Bucket reportBucket = storage.get(arguments.patientReportBucket());
         if (reportBucket == null) {
-            reportBucket = storage.create(BucketInfo.newBuilder(arguments.patientReportBucket())
+            BucketInfo.Builder builder = BucketInfo.newBuilder(arguments.patientReportBucket())
                     .setStorageClass(StorageClass.REGIONAL)
-                    .setLocation(arguments.region())
-                    .build());
+                    .setLocation(arguments.region());
+            if (arguments.cmek().trim().isEmpty()) {
```
pauldwolfe, line 31: I don't think its actually possible for cmek to be empty, its a mandatory argument now.

nedleitch: It can be set to a string of spaces (and maybe the empty string too, I didn't check that) using quotes on the command line.

PR #179, opened 2020-07-30
DEV-1428 Wrap all list operations in a backoff retry

pauldwolfe:
Now all interactions in the Lifecycle manager are retried with a backoff.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/execution/vm/InstanceLifecycleManager.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/execution/vm/InstanceLifecycleManager.java
@@ -5,8 +5,10 @@ import static java.lang.String.format;
 import java.io.File;
 import java.io.IOException;
 import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Optional;
+import java.util.Random;
 import java.util.stream.Collectors;
 
 import com.google.api.services.compute.Compute;
@@ -28,6 +30,7 @@ import net.jodah.failsafe.function.CheckedSupplier;
 class InstanceLifecycleManager {
     private static final String RUNNING_STATUS = "RUNNING";
     private static final Logger LOGGER = LoggerFactory.getLogger(InstanceLifecycleManager.class);
+    public static final int FIVE_MINUTES = 600;
 
     private final String project;
     private final Compute compute;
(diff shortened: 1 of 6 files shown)


cluster/src/main/java/com/hartwig/pipeline/execution/vm/InstanceLifecycleManager.java
```diff
@@ -128,21 +131,23 @@ private Operation executeSynchronously(ComputeRequest<Operation> request, String
...
-        return executeWithRetries(() -> compute.zoneOperations().get(projectName, zoneName, asyncOp.getName()).execute());
+        return executeWithRetries(() -> compute.zoneOperations().get(projectName, zoneName, asyncOp.getName()).execute(),
+                IOException.class);
     }
 
-    private Operation executeWithRetries(final CheckedSupplier<Operation> operationCheckedSupplier) {
-        return Failsafe.with(new RetryPolicy<>().handle(IOException.class).withDelay(Duration.ofSeconds(pollInterval)).withMaxRetries(5))
+    private <T> T executeWithRetries(final CheckedSupplier<T> operationCheckedSupplier, final Class<? extends Exception> exception) {
+        return Failsafe.with(new RetryPolicy<>().handle(exception)
+                .withBackoff(pollInterval, FIVE_MINUTES, ChronoUnit.SECONDS, new Random().nextInt(pollInterval)))
```
nedleitch, line 140: I wonder in retrospect whether the three-argument version of `withDelay` might be more appropriate (provide a min and max delay and it will choose a random time between them). I think it's more readable too (and you could do without the constant - just use `5` and specify minutes instead of seconds).

cluster/src/main/java/com/hartwig/pipeline/execution/vm/InstanceLifecycleManager.java
```diff
@@ -28,6 +30,7 @@
 class InstanceLifecycleManager {
     private static final String RUNNING_STATUS = "RUNNING";
     private static final Logger LOGGER = LoggerFactory.getLogger(InstanceLifecycleManager.class);
+    public static final int FIVE_MINUTES = 600;
```
nedleitch, line 33: This is 10 minutes :)

pauldwolfe: toucher

cluster/src/main/java/com/hartwig/pipeline/execution/vm/InstanceLifecycleManager.java
```diff
@@ -73,22 +75,22 @@ Operation deleteOldInstancesAndStart(Instance instance, String zone, String vmNa
...
-            return executeSynchronously(compute.instances().insert(project, zone, instance), project, zone);
+            return executeSynchronously(compute.instances().insert(project, zone, instance), project, zone, "insertVm");
         } catch (IOException ioe) {
             throw new RuntimeException("Could not initialise insert operation!", ioe);
         }
     }
 
     void delete(String zone, String vm) {
-        executeSynchronously(getWithRetries(() -> compute.instances().delete(project, zone, vm)), project, zone);
+        executeSynchronously(getWithRetries(() -> compute.instances().delete(project, zone, vm)), project, zone, "deleteVm");
```
nedleitch, line 85: Repeated op name, maybe put in a constant?

PR #181, opened 2020-08-06
DEV-1443 Create a profile for platinum Arguments

pauldwolfe:
The profile is called PUBLIC, which in this context means it is a profile which can
be shared publicly. Of course the intention it will always be invoked via platinum,
but maybe that won't always be the case.

First clean up all unused arguments and classes which still reference them. Also rename
private network to just network (private network isn't really a thing) and patient
report bucket to output bucket (no reports are in that bucket).

CMEK becomes an optional again, and the strict checking is disabled.

The public image is just enabled by a flag, which is actually not exposed to the CLI
and can only be enabled with the PUBLIC profile. This way we can never have public
mode running with a private image


cluster/src/main/java/com/hartwig/pipeline/Arguments.java
```diff
@@ -12,6 +12,7 @@
     String EMPTY = "";
 
     enum DefaultsProfile {
+        PUBLIC,
```
nedleitch, line 15: I'm wrestling with whether it would be more obvious if this were called `PLATINUM`.

cluster/src/main/java/com/hartwig/pipeline/alignment/AlignerProvider.java
```diff
@@ -77,9 +76,9 @@ private LocalAlignerProvider(final GoogleCredentials credentials, final Storage
         @Override
         VmAligner wireUp(GoogleCredentials credentials, Storage storage, AlignmentOutputStorage alignmentOutputStorage,
```
nedleitch, line 77: In the spirit of cleanup herein, would now be a good time to rename `VmAligner` to `Aligner`?

cluster/src/main/java/com/hartwig/pipeline/Arguments.java
pauldwolfe: the public profile can be used outside platinum.

cluster/src/main/java/com/hartwig/pipeline/alignment/AlignerProvider.java
pauldwolfe: should be a separate commit. The reason to tie up the argument cleanups with this one is that to make a new profile, you need to populate all those unnused args. Then the rest of the deletes are all cascading from deleting those arguments.

PR #182, opened 2020-08-06
Dev 1432: Use GRIPSS instead of somatic filter

jonbaber:
After Ned's feedback I have removed run_sage_caller argument and retained run_somatic_caller and run_structural_caller arguments.


cluster/src/main/java/com/hartwig/pipeline/calling/structural/gridss/stage/Driver.java
```diff
@@ -21,21 +23,20 @@
...
         super("gridss.driver", OutputFile.GZIPPED_VCF);
         this.assemblyBamPath = assemblyBamPath;
-        this.referenceGenomePath = referenceGenomePath;
-        this.blacklistBedPath = blacklistBedPath;
-        this.gridssConfigPath = gridssConfigPath;
-        this.repeatMaskerDbBed = repeatMaskerDbBed;
+        this.referenceGenomePath =  resourceFiles.refGenomeFile();
+        this.blacklistBedPath = resourceFiles.gridssBlacklistBed();
+        this.gridssConfigPath = resourceFiles.gridssPropertiesFile();
+        this.repeatMaskerDbBed = resourceFiles.gridssRepeatMaskerDbBed();
```
nedleitch, line 32: I'd be inclined to just store `resourceFiles` as an instance variable, the method names on it are descriptive enough.

PR #190, opened 2020-09-02
DEV-1481 Specify image from CLI

nedleitch:


cluster/src/main/java/com/hartwig/pipeline/CommandLineOptions.java
```diff
@@ -59,7 +63,7 @@
     private static final String SET_ID_FLAG = "set_id";
     private static final String SBP_RUN_ID_FLAG = "sbp_run_id";
     private static final String NETWORK_FLAG = "network";
-    private static final String CMEK_FLAG = "cmek";
+//    private static final String CMEK_FLAG = "cmek";
```
pauldwolfe, line 66: I think we should keep all flags used in the CLI in this class for now or do a larger refactor. Prefer the consistency to the small duplication between CommonArguments and this list.

cluster/src/main/java/com/hartwig/pipeline/CommandLineOptions.java
```diff
@@ -118,21 +122,26 @@ private static Options options() {
...
+                .addOption(json())
+                .addOption(imageName());
     }
 
     private static Option json() {
         return optionWithArg(SAMPLE_JSON_FLAG, "JSON file defining the location of FASTQ inputs in GCP.");
     }
 
+    private static Option imageName() {
+        return optionWithArg(IMAGE_NAME, "Image to use instead of the default");
```
pauldwolfe, line 134: lets make this a bit more descriptive

cluster/src/main/java/com/hartwig/pipeline/Arguments.java
```diff
@@ -161,11 +161,11 @@ static String workingDir() {
                     .publishToTurquoise(false)
                     .pollInterval(DEFAULT_POLL_INTERVAL)
                     .refGenomeVersion(DEFAULT_REF_GENOME_VERSION)
-                    .maxConcurrentLanes(DEFAULT_MAX_CONCURRENT_LANES);
+                    .maxConcurrentLanes(DEFAULT_MAX_CONCURRENT_LANES)
+                    .imageName(EMPTY);
```
pauldwolfe, line 165: As the argument is optional, we should use the features of Optional and not have another "empty" case of an empty string. You can either just not specify the image name in the builder or use Optional.empty(). I think the former is most consistent with what's already here.

batch/src/main/java/com/hartwig/batch/BatchArguments.java
```diff
@@ -45,7 +47,7 @@ static BatchArguments from(String[] args) {
                     .outputBucket(commandLine.getOptionValue(OUTPUT_BUCKET))
                     .cmek(commandLine.getOptionValue(CMEK, CommonArguments.DEFAULT_DEVELOPMENT_CMEK))
                     .network(commandLine.getOptionValue(PRIVATE_NETWORK, DEFAULT_NETWORK))
-                    .usePublicImage(false)
+                    .imageName(commandLine.getOptionValue(IMAGE_NAME, EMPTY))
```
pauldwolfe, line 50: I think better to just not expose the image name argument to batch, then you can also remove the check for empty string in the compute engine.

batch/src/test/java/com/hartwig/batch/input/InputParserProviderTest.java
```diff
@@ -16,17 +15,23 @@
...
         when(descriptor.inputType()).thenReturn(OperationDescriptor.InputType.JSON);
-        victim.from(operation);
-        assertThat(1).isEqualTo(1);
+        InputParser parser = victim.from(operation);
+        assertThat(parser.getClass()).isEqualTo(JsonInputParser.class);
     }
 
     @Test
     public void shouldReturnFlatParserWhenOperationSpecifiesIt() {
+        InputParserProvider victim = new InputParserProvider();
```
pauldwolfe, line 28: How is this related?

pauldwolfe (approved):

PR #191, opened 2020-09-02
DEV-1478 Run from persisted data given a starting point

pauldwolfe:
Adds the ability to specify a starting point to the pipeline. The starting point can be
either alignment, calling or purple complete. Each starting point builds off the last,
so purple complete is the union of all 3.

For each stage which can use persistent data, there is a PersistedXxx implementation.
These classes contain the knowledge of the location of the output files for outputs
created with this same version of the pipeline.

It works by using the exists "skipping" mechanism, but enriching the skipped output
with the actual location in google storage.

The remaining "OutputStorage" classes which used to poll for metrics and germline data
from the single sample pipelines are replaced with queues to avoid any duplication with
the persisted implementations.

The SomaticPipeline is also changed to take an AlignmentPair as input, rather than looking it up via the
storage. This allows us to pass either the "live" alignment or a stored one without any awareness
in the pipeline.

Cram is also supported, but still need to test. Not sure if all tools can work directly with a cram
without additional arguments.


cluster/src/main/java/com/hartwig/pipeline/Arguments.java
```diff
@@ -11,6 +11,8 @@
 
     String EMPTY = "";
 
+    Optional<String> runFrom();
```
nedleitch, line 14: Feels to me like `run` is a bit overloaded already. Could we use another verb? Like `begin` or `start` or `restart`?

cluster/src/test/java/com/hartwig/pipeline/SingleSamplePipelineTest.java
```diff
@@ -1,62 +1,18 @@
...
-import com.hartwig.pipeline.stages.StageRunner;
-
-import org.junit.Before;
-import org.junit.Test;
-import org.mockito.ArgumentCaptor;
-
 public class SingleSamplePipelineTest {
 
-    private static final Arguments ARGUMENTS = Arguments.testDefaults();
+/*    private static final Arguments ARGUMENTS = Arguments.testDefaults();
```
nedleitch, line 5: Could we have found a case where a `TODO` annotation would actually be worthwhile?

nedleitch (approved):

PR #198, opened 2020-09-17
DEV-1515 Add a completion message summarizing the run

pauldwolfe:
INFO when success, ERROR when not.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/PipelineMain.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/PipelineMain.java
@@ -184,6 +184,6 @@ public class PipelineMain {
     }
 
     public static String completionMessage(final PipelineState state) {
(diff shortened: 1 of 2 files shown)


cluster/src/main/java/com/hartwig/pipeline/PipelineMain.java
```diff
@@ -180,4 +182,8 @@ public static void main(String[] args) {
             System.exit(1);
         }
     }
+
+    public static String completionMessage(final PipelineState state) {
+        return String.format("Pipeline completed with status [%s], summary: [%s]", state, state.toString());
```
nedleitch, line 187: Don't both the `%s` templates have the same value?

nedleitch, line 187: Don't both the `%s` templates have the same value?

nedleitch (changes requested):

PR #199, opened 2020-09-21
DEV-1498 Use persisted data for somatic stages

pauldwolfe:
Refactors previous change to also support looking up somatic data from persistent storage.

By adding the barcode to the RunMetadata we can actually avoid having a different implementations
and look up the files using the tumor barcode.

As many of the HMF tools just need a directory, we add a second method to the dataset interface.
Maybe later we refactor the model in the API to better support this, so nice to hide it behind
and interface.

Add a simple check in the cleanup to ensure we don't fail deleting runtime buckets which do not
exist.


cluster/src/main/java/com/hartwig/pipeline/reruns/NoopPersistedDataset.java
```diff
@@ -0,0 +1,18 @@
+package com.hartwig.pipeline.reruns;
+
+import java.util.Optional;
+
+import com.hartwig.pipeline.datatypes.DataType;
+import com.hartwig.pipeline.metadata.RunMetadata;
+
+public class NoopPersistedDataset implements PersistedDataset {
```
nedleitch, line 8: Maybe it's just a nit but I do think this would be easier to read as `NoOpPersistedDataset`.

nedleitch (approved):

PR #200, opened 2020-09-22
DEV-1525 Publish datatypes for all files used in re-runs

pauldwolfe:
Generally follows the pattern laid out for crams, with a few changes.

Linking to sample is combined with datatype linking, we always need a sample to look it up.

Path construction done in the AddDatatypeToFile class to prevent mistakes and duplication.

Test added, still WIP.


cluster/src/main/java/com/hartwig/pipeline/sbpapi/SbpRestApi.java
```diff
@@ -113,10 +115,19 @@ public AddFileApiResponse postFile(final SbpFileMetadata metaData) {
...
+    public void linkFileToSample(final int id, final String barcode) {
+        try {
+            SbpSample sample = ObjectMappers.get().<List<SbpSample>>readValue(returnOrThrow(api().path(SAMPLES)
+                    .queryParam("barcode", barcode)
+                    .request()
+                    .get()), new TypeReference<List<SbpSample>>() {
+            }).stream().findFirst().orElseThrow();
+            Map<String, Integer> request = new HashMap<>();
+            request.put("sample_id", sample.id());
+            returnOrThrow(submitJson(POST, api().path(FILES).path(String.valueOf(id)).path("sample"), request, Response.Status.CREATED));
```
nedleitch, line 127: This is an opportunity to loosen the allowable response codes so we can make the API return a more standard response.

nedleitch (approved):

PR #202, opened 2020-09-25
Added VIRUSBreakend dependencies

d-cameron:
Added Kraken2 and RepeatMasker dependencies

- GRIDS 2.10.0 is pre-release. Not sure what we want to do about this. Not merge until I formally release it, or just update common-tools when it does get released.


build/repeatmasker.sh
```diff
@@ -0,0 +1,36 @@
+sudo apt-get -y install perl-modules cpanminus python3-h5py
```
nedleitch, line 1: Can you move this to `cluster/images/build/repeatmasker.sh` please? Top-level directories should be added judiciously and I feel as though having it nearer the rest of the scripts will help make it clear what it's for.

Also maybe just some comments at the top to indicate what it's for OR an amendment to the `cluster/images/README.md` file so people know they need to run it when the OS/major dependencies change but not on a regular basis, would be useful.

nedleitch (changes requested):

build/repeatmasker.sh
d-cameron: Added to cluster/images/build subdirectory. This still keeps it seperate from the image building scripts, but doesn't introduce a new top-level directory

nedleitch (approved):
Looks good, thanks Daniel! I'll leave it open in case there's anything else you want to do but feel free to merge.

pauldwolfe:
Closing this PR but leaving the branch in place.

PR #208, opened 2020-10-06
DEV-1543 Upgrade to chord and integrate with HG38

pauldwolfe:
Bump version and allow for passing of ref genome to the R script.


cluster/images/pipeline5.cmds
```diff
@@ -34,6 +34,7 @@ sudo chmod a+x /opt/tools/bwa/0.7.17/bwa
 sudo chmod a+x /opt/tools/sambamba/0.6.5/sambamba
 sudo chmod a+x /opt/tools/sambamba/0.6.8/sambamba
 sudo chmod a+x /opt/tools/chord/60.02_1.03/extractSigPredictHRD.R
+sudo chmod a+x /opt/tools/chord/2.00_1.14/extractSigPredictHRD.R
```
nedleitch, line 37: With the wonky versioning might it make sense to remove the `60.02_1.03` version now so it doesn't get kept instead of the new one in a later cleanup?

nedleitch (approved):

cluster/images/pipeline5.cmds
pauldwolfe: should be fine, i don't think we'd ever remove a version solely based on increments

PR #210, opened 2020-10-08
DEV-1555 Print failure logs at the end of pipeline

pauldwolfe:
Adds a field to the stage outputs to cache a reference to the log blob in case of failure.

Little awkward as we need to explicitly add the log everywhere, but can't find another solution
with the current state of the StageOutputs. The lanes in the BwaAligner also make it a bit
tricky to make a really general solution.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/execution/vm/BashStartupScript.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/execution/vm/BashStartupScript.java
@@ -36,7 +36,8 @@ public class BashStartupScript {
     }
 
     String asUnixString() {
-        return asUnixString(new StorageStrategy() {});
+        return asUnixString(new StorageStrategy() {
+        });
     }
 
     String asUnixString(StorageStrategy storageStrategy) {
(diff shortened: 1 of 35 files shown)


cluster/src/main/java/com/hartwig/pipeline/report/FailureSummary.java
```diff
@@ -0,0 +1,38 @@
...
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.Storage;
+import com.hartwig.pipeline.PipelineState;
+import com.hartwig.pipeline.StageOutput;
+import com.hartwig.pipeline.execution.PipelineStatus;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class FailureSummary {
```
nedleitch, line 15: I can see the reasoning (it's a summary showing any failures) but it's not necessarily a failure. Could we call it `StageSummary` or `ExecutionSummary` or something similar?

cluster/src/main/java/com/hartwig/pipeline/execution/vm/BashStartupScript.java
```diff
@@ -50,7 +50,7 @@ String asUnixString(StorageStrategy storageStrategy) {
                 "function die() {",
                 "  exit_code=$?",
                 "  echo \"Unknown failure: called command returned $exit_code\"",
-                format("  gsutil -m cp %s gs://%s", localLogFile, runtimeBucketName),
+                format("  gsutil -m cp %s gs://%s/results/", localLogFile, runtimeBucketName),
```
nedleitch, line 53: I thought we had a requirement that it live in the top-level?

cluster/src/main/java/com/hartwig/pipeline/resource/OverriddenReferenceGenome.java
```diff
@@ -53,7 +53,7 @@ public String gridssBreakendPon() {
 
     @Override
     public String gridssBreakpointPon() {
-        return decorated.gridssBreakpointPon();
+        return decorated.gridssBreakendPon();
```
nedleitch, line 56: How is this related?

nedleitch (changes requested):

cluster/src/main/java/com/hartwig/pipeline/report/FailureSummary.java
pauldwolfe: See your point, going with VmExecutionLogSummary

cluster/src/main/java/com/hartwig/pipeline/resource/OverriddenReferenceGenome.java
pauldwolfe: Not, was used to introduce a failure

cluster/src/main/java/com/hartwig/pipeline/execution/vm/BashStartupScript.java
pauldwolfe: don't think so, when the run succeeds this is where the logs go.

PR #213, opened 2020-10-17
DEV-1575 Add command line options for subnet and network tags

pauldwolfe:
Subnet is optional, and if not specified the old behaviour of using the same name as the
network is used (and the region for everything else). Tags are added to the instance when
specified.

### DIFF
--- a/cluster/src/test/java/com/hartwig/pipeline/execution/vm/GoogleComputeEngineTest.java
+++ b/cluster/src/test/java/com/hartwig/pipeline/execution/vm/GoogleComputeEngineTest.java
@@ -225,6 +225,25 @@ public class GoogleComputeEngineTest {
     }
 
     @Test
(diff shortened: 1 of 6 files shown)


cluster/src/test/java/com/hartwig/pipeline/execution/vm/GoogleComputeEngineTest.java
```diff
@@ -206,9 +207,9 @@ public void usesPublicNetworkIfNoPrivateSpecified() {
     }
 
     @Test
-    public void usesPrivateNetworkWhenSpecified() {
+    public void usesNetworkAndSubnetWhenSpecified() {
```
nedleitch, line 210: Should we also have a test to ensure it defaults to the network when the subnet isn't given?

nedleitch (approved):

PR #225, opened 2020-11-11
DEV-1645 Automate tool upgrade, image creation and docker image

pauldwolfe:
Simple script to copy the new jar into the common-tools bucket and start the image creation process. This now
is broken up into two sets of commands: base and tools. Passing the -t flag will create an image from the
previous build in the family, only overlaying the tools dir.

Otherwise the upgrade script does a maven build to create a docker container and push it to GCR (rather than
docker hub).


cluster/images/generate_imaging_script.sh
```diff
@@ -4,29 +4,41 @@
...
+        f) FLAVOUR=${OPTARG} ;;
+        *) ;;
+    esac
+done
+sourceInstance="pipeline5-${VERSION}${FLAVOUR+"-$FLAVOUR"}"
+image_project="debian-cloud"
+image_family="debian-9"
+base_image_cmds="$(dirname "$0")/base.cmds"
+tools_image_cmds="$(dirname "$0")/tools.cmds"
+all_cmds=$(echo $base_image_cmds $tools_image_cmds)
```
nedleitch, line 23: This can be written as `all_cmds="$base_image_cmds $tools_image_cmds"`

cluster/upgrade_tool.sh
```diff
@@ -0,0 +1,19 @@
...
+full=${tool}-${version}
+local_jar="${full}.jar"
+if [ -n "$3" ]; then
+  local_jar="$3"
+else
+  wget "https://github.com/hartwigmedical/hmftools/releases/download/${tool}-v${version}/${full}.jar"
+fi
+gsutil cp $local_jar "gs://common-tools/${tool}/${version}/${tool}.jar"
+if [ -z "$3" ]; then
+  rm local_jar
```
nedleitch, line 12: Missing `$`

cluster/upgrade_tool.sh
```diff
@@ -0,0 +1,19 @@
+tool=$1
```
nedleitch, line 1: Recommend starting with this preamble given how many things could potentially fail in this script:

```
#!/usr/bin/env bash

set -e
```
That way it will die straight away if say the `wget` fails.

cluster/upgrade_tool.sh
```diff
@@ -0,0 +1,19 @@
...
+  local_jar="$3"
+else
+  wget "https://github.com/hartwigmedical/hmftools/releases/download/${tool}-v${version}/${full}.jar"
+fi
+gsutil cp $local_jar "gs://common-tools/${tool}/${version}/${tool}.jar"
+if [ -z "$3" ]; then
+  rm local_jar
+fi
+
+./images/create_custom_image.sh -t
```
nedleitch, line 15: Would prefer something a little more robust here, like `"$(dirname $0)/images/create_custom_image.sh -t"` so we've made a reasonable effort to allow it to be run from outside the directory it lives in.

nedleitch (changes requested):

cluster/upgrade_tool.sh
pauldwolfe: it needs to run in the directory of the poms

PR #227, opened 2020-11-15
DEV-1656 Add argument to override the image project

pauldwolfe:


nedleitch (changes requested):
I don't think the addition of the `cluster/cobalt-1-11.jar.1` file could have been part of this ticket. I can't comment on the file itself.

Otherwise looks good.

PR #230, opened 2020-11-24
DEV-1677: Sage Germline Post Processing

jonbaber:


batch-operations/src/main/java/com/hartwig/batch/operations/SageGermline.java
```diff
@@ -31,73 +29,66 @@
...
 
     @Override
     public OperationDescriptor descriptor() {
-        return OperationDescriptor.of("SageGermline", "Generate germline output", OperationDescriptor.InputType.JSON);
+        return OperationDescriptor.of("SageGermline", "Generate sage output", OperationDescriptor.InputType.JSON);
+    }
+
+    private static String localFilename(InputFileDescriptor remote) {
+        return format("%s/%s", VmDirectories.INPUT, new File(remote.inputValue()).getName());
+    }
```
nedleitch, line 87: We already have the `localDestination` method on `InputFileDescriptor` that does this I think.

cluster/src/main/java/com/hartwig/pipeline/calling/command/BcfToolsCommandListBuilder.java
```diff
@@ -92,4 +107,15 @@ private void addArguments(String... argumentArray) {
...
     }
+
+    private static String singleQuote(String expression) {
+        String trimmed = expression.trim();
+
+        if (!trimmed.startsWith("'")) {
+            return "'" + expression + (trimmed.endsWith("'") ? "" : "'");
+        }
+
+        return expression + (trimmed.endsWith("'") ? "" : "'");
```
nedleitch, line 118: My fondness for regular expressions has its detractors but in this case I think something like this makes for a tidier implementation:

`return String.format("'%s'", expression.trim().replaceAll("^'", "").replaceAll("'$", ""));`

cluster/src/main/java/com/hartwig/pipeline/calling/somatic/BlacklistFilter.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public static final String BLACKLIST_FILTER = "BLACKLIST";
+
+    BlacklistFilter() {
+        super("blacklist.filter", FileTypes.GZIPPED_VCF);
+    }
+
+    @Override
+    public List<BashCommand> bash(final OutputFile input, final OutputFile output) {
+        final String expression = BlacklistBedAnnotation.BLACKLIST_BED_FLAG + "=1 || " + BlacklistVcfAnnotation.BLACKLIST_VCF_FLAG + "=1";
```
nedleitch, line 21: Nit: `String.format` might make this easier to read.

nedleitch (approved):

batch-operations/src/main/java/com/hartwig/batch/operations/SageGermline.java
jonbaber: Thanks. I was the one who put it there :)

cluster/src/main/java/com/hartwig/pipeline/calling/command/BcfToolsCommandListBuilder.java
jonbaber: That is much nicer. I have always avoided regular expressions in Java because apparently it is slow but of course it doesn't matter here. Nice tip. Thanks.

cluster/src/main/java/com/hartwig/pipeline/calling/somatic/BlacklistFilter.java
jonbaber: agreed

PR #233, opened 2020-12-07
DEV-1701 - Renamed HG19 to 37 and HG38 to 38 as used for resource files

charlesshale:


cluster/src/main/java/com/hartwig/pipeline/resource/RefGenomeVersion.java
```diff
@@ -1,6 +1,6 @@
 package com.hartwig.pipeline.resource;
 
 public enum RefGenomeVersion {
-    HG19,
-    HG38
+    RG_37,
```
pauldwolfe, line 4: I guess the "RG" is introduced as an enum value needs to start with a letter. Is this a convention you've seen elsewhere? RG is also used as an abbreviation for Read Group.

An alternative could  be `V37`/`V38`?

charlesshale: I don't mind and am fine just leaving it as-is. Dont think Read Group is a particularly common term so no issue there.

pauldwolfe: Ok I'll merge it as is and get a second opinion from Stef/Korneel.

PR #235, opened 2020-12-16
Dev 1723 - Sage Germline Calling

jonbaber:


cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageGermlineCaller.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public VirtualMachineJobDefinition vmDefinition(final BashStartupScript bash, final ResultsDirectory resultsDirectory) {
+        return VirtualMachineJobDefinition.sageGermlineCalling(bash, resultsDirectory);
+    }
+
+    @Override
+    public boolean shouldRun(final Arguments arguments) {
+        return arguments.runGermlineCaller();
```
pauldwolfe, line 69: Can we add a new toggle here so we can turn it off without disabling the GATK germline calling?

cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageSomaticCaller.java
```diff
@@ -0,0 +1,96 @@
...
+import com.hartwig.pipeline.execution.vm.unix.UnzipToDirectoryCommand;
+import com.hartwig.pipeline.metadata.SomaticRunMetadata;
+import com.hartwig.pipeline.reruns.PersistedDataset;
+import com.hartwig.pipeline.resource.ResourceFiles;
+import com.hartwig.pipeline.stages.SubStageInputOutput;
+import com.hartwig.pipeline.storage.RuntimeBucket;
+
+public class SageSomaticCaller extends SageCaller {
+
+    public static final String NAMESPACE = "sage_somatic";
```
pauldwolfe, line 25: What impact will this have downstream on reporting, etc? Will we need to do another rerun to make data homogenous again? 

We could limit this impact of this change by not changing this namespace yet and making that a separate ticket.

cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageCaller.java
```diff
@@ -0,0 +1,101 @@
...
+                .addFurtherOperations(new AddDatatypeToFile(dataType, Folder.root(), namespace(), filteredOutputFile, metadata.barcode()));
+    }
+
+    @Override
+    public SageOutput skippedOutput(final SomaticRunMetadata metadata) {
+        return SageOutput.builder(namespace()).status(PipelineStatus.SKIPPED).build();
+    }
+
+    @Override
+    public final SageOutput persistedOutput(final SomaticRunMetadata metadata) {
```
pauldwolfe, line 77: Maybe this guy is better only to implement in the somatic caller. Generally "persisted output" is only required for stages with other stages which depend on them, so you could leave this out of sage germline.

pauldwolfe (changes requested):

cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageGermlineCaller.java
jonbaber: Done. Added run_sage_germline_caller argument.

cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageSomaticCaller.java
jonbaber: Done. Reverted to sage.

cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageCaller.java
jonbaber: Left as discussed in email thread as we will be running GATK and sage germline in parallel for a while.

PR #236, opened 2020-12-21
DEV-1731 Add fields to ref genome enum for different representations

pauldwolfe:
One field per use case:linx; sage; chord; the resources; and the pipeline cli.

The enum values are renamed to V37 and V38 for simplicity.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/resource/RefGenomeVersion.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/resource/RefGenomeVersion.java
@@ -1,6 +1,40 @@
 package com.hartwig.pipeline.resource;
 
 public enum RefGenomeVersion {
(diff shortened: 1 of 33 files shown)


cluster/src/main/java/com/hartwig/pipeline/resource/RefGenomeVersion.java
```diff
@@ -1,6 +1,40 @@
 package com.hartwig.pipeline.resource;
 
 public enum RefGenomeVersion {
-    RG_37,
-    RG_38
+    V37("hg19", "HG37", "37", "37", "HG19"),
+    V38("hg38", "HG38", "38", "38", "HG38");
```
nedleitch, line 5: Nice solution.

nedleitch (approved):

PR #237, opened 2020-12-21
DEV-1732 Publish run result along with status

pauldwolfe:
Map the current status to the correct status + run result combination.


cluster/src/main/java/com/hartwig/pipeline/sbpapi/SbpRunResult.java
```diff
@@ -0,0 +1,37 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.hartwig.pipeline.execution.PipelineStatus;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonDeserialize(as = ImmutableSbpRunResult.class)
+@JsonSerialize(as = ImmutableSbpRunResult.class)
+public interface SbpRunResult {
```
nedleitch, line 12: While it might be inconsistent, this would also be an opportunity to not add more SBP legacy.

nedleitch (approved):

PR #252, opened 2021-02-22
DEV-1815 Add protect to the pipeline

pauldwolfe:
Protect is relatively simple but has a lot of dependencies, so most of this PR is wiring those up
(also for reruns).

Otherwise protect will now be the last stage of the pipeline.


cluster/src/main/java/com/hartwig/pipeline/tertiary/protect/Protect.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @Override
+    public VirtualMachineJobDefinition vmDefinition(final BashStartupScript bash, final ResultsDirectory resultsDirectory) {
+        return ImmutableVirtualMachineJobDefinition.builder()
+                .name(namespace())
+                .startupCommand(bash)
+                .namespacedResults(resultsDirectory)
+                .performanceProfile(VirtualMachinePerformanceProfile.custom(4, 12))
+                .build();
```
nedleitch, line 106: Think we talked some time ago about revising the disk requirements downward for post-alignment steps, not sure if this is a candidate? Looks like a modest VM otherwise so maybe it doesn't take long to run anyway?

cluster/src/main/java/com/hartwig/pipeline/tertiary/protect/ProtectCommand.java
```diff
@@ -0,0 +1,52 @@
...
+import com.hartwig.pipeline.execution.vm.JavaJarCommand;
+import com.hartwig.pipeline.tools.Versions;
+
+public class ProtectCommand extends JavaJarCommand {
+
+    public ProtectCommand(final String tumorSample, final List<String> primaryTumorDoids, final String outputDir,
+            final String actionabilityDir, final String doidJsonPath, final String germlineReportingTsvPath, final String purplePurityPath,
+            final String purpleQCFilePath, final String purpleDriverCatalogPath, final String purpleSomaticVariantsPath,
+            final String bachelorTsvPath, final String linxFusionTsvPath, final String linxBreakendTsvPath, final String linxDriversTsvPath,
+            final String linxViralInsertionsTsvPath, final String chordPredictionPath) {
```
nedleitch, line 14: This has to be some sort of record for parameter count.

Could we remove the output directory as it looks like it's a constant?

cluster/src/test/java/com/hartwig/pipeline/tertiary/linx/LinxTest.java
```diff
@@ -50,8 +50,20 @@ public void setUp() throws Exception {
     @Override
     protected List<ApiFileOperation> expectedFurtherOperations() {
         return List.of(new AddDatatype(DataType.LINX,
-                TestInputs.defaultSomaticRunMetadata().barcode(),
-                new ArchivePath(Folder.root(), Linx.NAMESPACE, "tumor.linx.breakend.tsv")));
+                        TestInputs.defaultSomaticRunMetadata().barcode(),
+                        new ArchivePath(Folder.root(), Linx.NAMESPACE, "tumor.linx.breakend.tsv")),
```
nedleitch, line 54: This filename seems wrong.

nedleitch (changes requested):

cluster/src/main/java/com/hartwig/pipeline/tools/Versions.java
```diff
@@ -29,6 +29,7 @@
     String CHORD = "2.00_1.14";
     String SAMTOOLS = "1.10";
     String BAMCOMP = "1.3";
+    String PROTECT = "";
```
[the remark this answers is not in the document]
pauldwolfe: already fixed

cluster/src/main/java/com/hartwig/pipeline/tertiary/protect/Protect.java
pauldwolfe: yeah good idea.

PR #254, opened 2021-03-01
DEV-1834 Publish staging events when pipeline complete in rerun mode

pauldwolfe:
Adds a new implementation of the complete method which publishes a PipelineStaged event when the pipeline has
not failed. This involves iterating through the output and passing each path to the event,along with its
optional data type. The datatypes are now linked in the event subscriber.

Due to this, it was practical to remove the interface which AddDatatype implemented (no other implementations).
Eventually will be able to simplify it further once live pipelines are archived this way.

Jackson version needed bumping to be compatible with the new event api.


cluster/src/main/java/com/hartwig/pipeline/metadata/BiopsyMetadataApi.java
```diff
@@ -62,6 +85,49 @@ public void start() {
...
+                    state.stageOutputs().stream().map(StageOutput::datatypes).flatMap(List::stream).collect(Collectors.toList());
+            SampleSet set = OnlyOne.of(setApi.list(metadata.set(), null), SampleSet.class);
+            ImmutablePipelineStaged.Builder stagedEventBuilder = ImmutablePipelineStaged.builder()
+                    .type(PipelineStaged.Type.DNA)
+                    .version(Versions.pipelineMajorMinorVersion())
+                    .sample(metadata.tumor().sampleName())
+                    .setId(set.getId());
+            OutputIterator.from(blob -> {
+                Optional<DataType> dataType =
+                        addDatatypes.stream().filter(d -> blob.getName().endsWith(d.path())).map(AddDatatype::dataType).findFirst();
```
nedleitch, line 99: Should we include `/` in the `endsWith` string as an anchor? eg `...endsWith("/" + d.path()))...`?

cluster/src/main/java/com/hartwig/pipeline/metadata/BiopsyMetadataApi.java
```diff
@@ -62,6 +85,49 @@ public void start() {
...
+                                .barcode(metadata.barcode())
+                                .bucket(blob.getBucket())
+                                .filesize(blob.getSize())
+                                .hash(MD5s.convertMd5ToSbpFormat(blob.getMd5()))).build());
+            }, sourceBucket).iterate(metadata);
+            stagedEventBuilder.build().publish(publisher, objectMapper);
+        }
+    }
+
+    private static ImmutablePipelineOutputBlob.Builder root(final String tumorSample, final String refSample, final String blobName,
```
nedleitch, line 114: Two things I don't like about this method, but for which I don't have great solutions:

* It mutates the state of the builder it is passed in, then returns it. In practice you're creating the builder inline in the caller so I understand why and also it maybe isn't a big problem, just don't love the way it reads.
* Don't love the name. Maybe `pathComponents` or `splitPath`? I dunno.

cluster/src/test/java/com/hartwig/pipeline/metadata/BiopsyMetadataApiTest.java
```diff
@@ -60,11 +123,106 @@ public void returnsMetadataForBiopsySamples() {
...
+        ArgumentCaptor<PubsubMessage> pubsubMessageArgumentCaptor = pipelineCompleteWithFile("set/purple/tumor.purple.somatic.vcf.gz",
+                TestOutput.builder().status(PipelineStatus.SUCCESS).build());
+
+        PipelineStaged result =
+                ObjectMappers.get().readValue(pubsubMessageArgumentCaptor.getValue().getData().toByteArray(), PipelineStaged.class);
+        assertThat(result.type()).isEqualTo(PipelineStaged.Type.DNA);
+        assertThat(result.runId()).isEmpty();
+        assertThat(result.setId()).isEqualTo(SET_ID);
+        assertThat(result.sample()).isEqualTo("tumor");
+        assertThat(result.version()).isEqualTo("local-SNAPSHOT");
```
nedleitch, line 137: Will this work in the build?

cluster/src/main/java/com/hartwig/pipeline/metadata/BiopsyMetadataApi.java
pauldwolfe: agree on both, i'll see what i can do to improve

cluster/src/test/java/com/hartwig/pipeline/metadata/BiopsyMetadataApiTest.java
pauldwolfe: prob not :)

cluster/src/main/java/com/hartwig/pipeline/metadata/MD5s.java
```diff
@@ -0,0 +1,12 @@
+package com.hartwig.pipeline.metadata;
+
+import java.util.Base64;
+
+import org.apache.commons.codec.binary.Hex;
+
+public class MD5s {
+
+    static String convertMd5ToSbpFormat(String originalMd5) {
```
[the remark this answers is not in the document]
pauldwolfe: yup

PR #259, opened 2021-03-15
DEV-1866: Bump PROTECT to 1.2

stefvanlieshout:
Updated PROTECT to 1.2 and cancerPanel testrun ran successful.


cluster/src/main/java/com/hartwig/pipeline/PipelineMain.java
```diff
@@ -46,7 +46,7 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(PipelineMain.class);
 
-    public PipelineState start(Arguments arguments) {
+    public PipelineState start(Arguments arguments) throws Exception {
```
pauldwolfe, line 49: Why does this method throw now?

pauldwolfe (approved):

cluster/src/main/java/com/hartwig/pipeline/PipelineMain.java
stefvanlieshout: That was unintentional so feel free to revert!

PR #262, opened 2021-03-16
DEV-1881 Convert crams to bams if used for downstream algos

pauldwolfe:
Adds an optional stage between the aligner and other tools which will convert any alignment with the
extension .cram back to .bam and use that downstream.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/cram2bam/Cram2Bam.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/cram2bam/Cram2Bam.java
@@ -1,0 +1,85 @@
+package com.hartwig.pipeline.cram2bam;
+
+import java.util.List;
+
+import com.hartwig.pipeline.Arguments;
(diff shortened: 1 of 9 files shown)


cluster/src/main/java/com/hartwig/pipeline/cram2bam/Cram2Bam.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public String namespace() {
+        return "cram2bam";
+    }
+
+    @Override
+    public List<BashCommand> commands(final SingleSampleRunMetadata metadata) {
+        String outputBam = VmDirectories.OUTPUT + "/" + FileTypes.bam(metadata.sampleName());
```
nedleitch, line 46: Prefer `VmDirectories.outputFile(...)`.

nedleitch (approved):

PR #269, opened 2021-04-15
DEV-1679 Add VIRUSBreakend

stefvanlieshout:
All is in place to run VirusBreakend but no doubt some things might need improvement. Let me know if that's the case!


cluster/images/tools.cmds
```diff
@@ -24,6 +28,17 @@ sudo chmod a+x /opt/tools/gridss/2.7.2/gridss.sh
 sudo chmod a+x /opt/tools/gridss/2.8.3/gridss.sh
 sudo chmod a+x /opt/tools/gridss/2.9.3/gridss.sh
 sudo chmod a+x /opt/tools/gridss/2.9.4/gridss.sh
+
```
pauldwolfe, line 31: For gridss 2.11 we only needed gridss.sh and gridss_annotate_vcf_repeatmasker.sh.

cluster/src/main/java/com/hartwig/pipeline/datatypes/DataType.java
```diff
@@ -21,6 +21,8 @@
     CHORD_PREDICTION,
     COBALT,
     AMBER,
+    VIRUSBREAKEND_SUMMARY,
```
pauldwolfe, line 24: We only need to add datatypes for files which will either be consumed by another step or persisted to the DB. Do we plan to make a virusbreakend loader for the DB which uses these files?

cluster/src/main/java/com/hartwig/pipeline/execution/vm/VirtualMachineJobDefinition.java
```diff
@@ -129,6 +129,14 @@ static VirtualMachineJobDefinition amber(BashStartupScript startupScript, Result
                 .build();
     }
 
+    static VirtualMachineJobDefinition virusbreakend(BashStartupScript startupScript, ResultsDirectory resultsDirectory) {
+        return ImmutableVirtualMachineJobDefinition.builder()
```
pauldwolfe, line 133: this will use the default machine size of 32 cpu 120G mem. Do we have any idea of the performance profile? GRIDSS uses 12/64.

cluster/src/main/java/com/hartwig/pipeline/tertiary/virusbreakend/VirusBreakend.java
```diff
@@ -0,0 +1,104 @@
...
+    @Override
+    public String namespace() {
+        return NAMESPACE;
+    }
+
+    @Override
+    public List<BashCommand> commands(final SomaticRunMetadata metadata) {
+        return List.of(
+                // TODO implement class to export PATH of latest version of tool
+                () -> "export PATH=/opt/tools/gridss/2.11.1:$PATH",
```
pauldwolfe, line 48: we already have a handy class actually, have a look at ExportPathCommand

cluster/src/main/java/com/hartwig/pipeline/tertiary/virusbreakend/VirusBreakendCommand.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.hartwig.pipeline.calling.command.VersionedToolCommand;
+import com.hartwig.pipeline.execution.vm.BashCommand;
+import com.hartwig.pipeline.execution.vm.OutputFile;
+import com.hartwig.pipeline.execution.vm.VmDirectories;
+import com.hartwig.pipeline.resource.ResourceFiles;
+import com.hartwig.pipeline.tools.Versions;
+import static com.hartwig.pipeline.resource.ResourceNames.VIRUSBREAKEND_DB;
+
+public class VirusBreakendCommand implements BashCommand {
```
pauldwolfe, line 11: VirusBreakendCommand can `extend` VersionToolCommand. Then you just pass the arguments to the super constructor. See `ChordExtractSigPredictHRD` for an example

cluster/src/main/java/com/hartwig/pipeline/tertiary/virusbreakend/VirusBreakendOutput.java
```diff
@@ -0,0 +1,27 @@
...
+
+@Value.Immutable
+public interface VirusBreakendOutput extends StageOutput {
+
+    @Override
+    default String name() {
+        return VirusBreakend.NAMESPACE;
+    }
+
+    Optional<GoogleStorageLocation> maybeOutputDirectory();
```
pauldwolfe, line 18: Does this get used by anything? If not fine to leave it out.

cluster/src/main/java/com/hartwig/pipeline/tools/Versions.java
```diff
@@ -31,6 +31,9 @@
     String BAMCOMP = "1.3";
     String PROTECT = "1.2";
 
+    // Version of VirusBreakend refers to GRIDSS version to use
+    String GRIDSS_VIRUS_BREAKEND = "2.11.1";
```
pauldwolfe, line 35: You'll be able to consolidate this now with GRIDSS when u merge

cluster/images/tools.cmds
stefvanlieshout: Ok removed everything else that is not used by VirusBreakend.

cluster/src/main/java/com/hartwig/pipeline/datatypes/DataType.java
stefvanlieshout: The summery contains information for report so could be used downstream in the future. And the VCF contains all breakpoints so could, and probably should, be loaded into patientDB at some point.

cluster/src/main/java/com/hartwig/pipeline/execution/vm/VirtualMachineJobDefinition.java
stefvanlieshout: No idea to be honest but I can't imagine VirusBreakend using more resources than GRIDSS, so I've now configured 12/64 as well.

cluster/src/main/java/com/hartwig/pipeline/tertiary/virusbreakend/VirusBreakend.java
stefvanlieshout: Sweet stuff, using it now ;)

cluster/src/main/java/com/hartwig/pipeline/tertiary/virusbreakend/VirusBreakendCommand.java
stefvanlieshout: Much cleaner indeed, thanks.

cluster/src/main/java/com/hartwig/pipeline/tools/Versions.java
stefvanlieshout: Might be better to leave it like this in case we want to update to the one or the other later on? Essentially these are two different tools, so specifying different version makes sense I think?

cluster/src/main/java/com/hartwig/pipeline/tertiary/virusbreakend/VirusBreakendCommand.java
```diff
@@ -0,0 +1,28 @@
...
+import static com.hartwig.pipeline.resource.ResourceNames.VIRUSBREAKEND_DB;
+
+public class VirusBreakendCommand extends VersionedToolCommand {
+
+    public VirusBreakendCommand(ResourceFiles resourceFiles, String tumorSample, String tumorBamPath) {
+        super("gridss",
+                "virusbreakend.sh",
+                Versions.GRIDSS_VIRUS_BREAKEND,
+                "--output",
+                VmDirectories.OUTPUT + "/" + tumorSample + ".virusbreakend.vcf",
```
nedleitch, line 16: There are some helper methods on `VmDirectories` to create paths.

nedleitch (approved):

cluster/src/main/java/com/hartwig/pipeline/execution/vm/VirtualMachineJobDefinition.java
stefvanlieshout: Discussed and will configure to 12 cores, 40 GB mem. Should cover it.

cluster/src/main/java/com/hartwig/pipeline/tertiary/virusbreakend/VirusBreakendOutput.java
stefvanlieshout: Not used atm indeed so will take it out.

cluster/src/main/java/com/hartwig/pipeline/tertiary/virusbreakend/VirusBreakendCommand.java
stefvanlieshout: Ok will use those indeed.

PR #273, opened 2021-04-21
DEV-1932 Cuppa main program and chart generation

nedleitch:
Call the main application which is a Java program like most of the
others. The chart is generated by a Python application called against
the output of the Java program.


cluster/src/main/java/com/hartwig/pipeline/Arguments.java
```diff
@@ -27,106 +27,6 @@
 
     boolean runBamMetrics();
 
-    boolean runSnpGenotyper();
```
pauldwolfe, line 30: has anything actually changed in here? would be nice to remove this diff

cluster/src/main/java/com/hartwig/pipeline/resource/ResourceFiles.java
```diff
@@ -110,6 +111,10 @@ default String gridssVirusRefGenomeFile() {
 
     String actionabilityDir();
 
+    default String cuppaRefData() {
+        return VmDirectories.resourcesPath(CUPPA);
```
pauldwolfe, line 115: can we make this more consistent with the `of` method used by the other resources?

cluster/src/main/java/com/hartwig/pipeline/tertiary/cuppa/Cuppa.java
```diff
@@ -0,0 +1,158 @@
...
+                                "-sample_data_dir",
+                                VmDirectories.INPUT,
+                                "-sample_sv_file",
+                                purpleStructuralVcfDownload.getLocalTargetPath(),
+                                "-sample_somatic_vcf",
+                                purpleSomaticVcfDownload.getLocalTargetPath(),
+                                "-log_debug",
+                                "-output_dir",
+                                VmDirectories.OUTPUT)),
+                () -> join(" ",
```
pauldwolfe, line 100: I think we should introduce a PythonCommand abstraction over using an anonymous lambda. Also looks strange to have the join rather than just making it a single list with two commands.

cluster/src/main/java/com/hartwig/pipeline/tertiary/cuppa/CuppaOutput.java
```diff
@@ -0,0 +1,26 @@
...
+    static ImmutableCuppaOutput.Builder builder() {
+        return ImmutableCuppaOutput.builder();
+    }
+
+    @Override
+    default String name() {
+        return Cuppa.NAMESPACE;
+    }
+
+    Optional<GoogleStorageLocation> maybeChart();
```
pauldwolfe, line 21: this can be removed as no downstream tools consume the chart

pauldwolfe (approved):

PR #275, opened 2021-04-23
DEV-1949 Add PEACH to the pipeline

nedleitch:
This change is mostly about reworking the way the Python applications are
packaged and run. PEACH needs Python 3.6, while the version in the OS is 3.5.
Because of this the imager now builds a virtual environment for PEACH, but it
will be easy to add another for Cuppa and any others that come along.

How this setup works is that we specify a tool name and version along
with a Python version in the imager script. An external script is invoked
by the imager to call `pyenv` to compile the target Python interpreter and
create a virtualenv. It then unpacks the given version of the tool and
installs its dependencies (from requirements.txt) in the new virtualenv.

I've also added a `Python3Command` for interfacing with all of this at
pipeline runtime. It takes a tool name and version
and generates shell code to call the tool with the environment built up
 by the imager for the tool.

 Two other unrelated changes are also in this commit: (1) an extra line in the
 imager commands to fix the broken sliced downloads for gsutil (look for
  "crcmod"), and (2) a slight refactoring of the Java-related BashCommands
into their own packages. Both of these were opportunistic and minor,
but clearly off-topic.


cluster/src/main/java/com/hartwig/pipeline/tertiary/peach/Peach.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    public String namespace() {
+        return NAMESPACE;
+    }
+
+    @Override
+    public List<BashCommand> commands(final SomaticRunMetadata metadata) {
+        return List.of(new Python3Command("peach",
+                Versions.PEACH,
+                "src/main.py",
```
pauldwolfe, line 50: perhaps we can ask him to create a wrapper script so we don't access a source file directly? (although im ok to ask for that in his next release)

pauldwolfe (approved):

PR #280, opened 2021-04-26
DEV-1965 Publish output target with stage events

pauldwolfe:
Depending on the context. The clinical API publishes the "patient report" target and the research api
publishes the "database" target.


nedleitch (approved):
Like the choice of terminology, really clarifies I think.

PR #292, opened 2021-05-18
DEV-2017: Override hotspot qual cutoff to 40 in Sage somatic caller w…

kduyvesteyn:
When we run SAGE in shallow mode we should reduce the cutoff for hotspots to retain roughly similar sensitivity. 

Analysis showed 40 is a good tradeoff between FP and FN/TP.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageCommandBuilder.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageCommandBuilder.java
@@ -143,14 +143,6 @@ public class SageCommandBuilder {
             throw new IllegalStateException("Shallow somatic mode enabled while not in shallow mode");
         }
 
-        if (somaticMode && germlineMode) {
-            throw new IllegalStateException("Somatic mode and germline mode both set while mutually exclusive");
-        }
-
-        if (!somaticMode && !germlineMode) {
-            throw new IllegalStateException("Neither somatic mode nor germline set while either needs to be enabled");
-        }
-
         final String tumorBamFiles = String.join(",", tumorBam);
         arguments.add("-tumor").add(tumor.toString()).add("-tumor_bam").add(tumorBamFiles);
         if (reference.length() > 0) {
@@ -164,9 +156,10 @@ public class SageCommandBuilder {
             if (shallowSomaticMode) {
                 arguments.add("-hotspot_min_tumor_qual").add("40");
             }
-        }
+        } else {
+            // Somatic mode and germline mode are mutually exclusive and this can't be controlled from outside this class.
+            assert germlineMode;
 
-        if (germlineMode) {
             arguments.add("-hotspots").add(resourceFiles.sageGermlineHotspots());
             arguments.add("-panel_bed").add(resourceFiles.sageGermlineCodingPanel());
             arguments.add("-hotspot_min_tumor_qual").add("50");
--- a/cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageCommandBuilder.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageCommandBuilder.java
@@ -23,6 +23,7 @@ public class SageCommandBuilder {
     private boolean ponMode = false;
     private boolean somaticMode = true;
     private boolean germlineMode = false;
+    private boolean shallowSomaticMode = false;
     private int tumorSamples = 0;
 
     public SageCommandBuilder(ResourceFiles resourceFiles) {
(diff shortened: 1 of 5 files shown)


cluster/src/test/java/com/hartwig/pipeline/calling/sage/SageSomaticCallerTest.java
```diff
@@ -24,7 +24,7 @@ public void setUp() throws Exception {
 
     @Override
     protected Stage<SageOutput, SomaticRunMetadata> createVictim() {
-        return new SageSomaticCaller(TestInputs.defaultPair(), TestInputs.REF_GENOME_37_RESOURCE_FILES, new NoopPersistedDataset());
+        return new SageSomaticCaller(TestInputs.defaultPair(), TestInputs.REF_GENOME_37_RESOURCE_FILES, new NoopPersistedDataset(), false);
     }
```
pauldwolfe, line 29: Can we add a test for this option? Something like PurpleTest.shallowModeUsesLowDepthSettings()

pauldwolfe (changes requested):

cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageCommandBuilder.java
```diff
@@ -68,6 +69,11 @@ public SageCommandBuilder addCoverage() {
         return this;
     }
 
+    public SageCommandBuilder addShallowSomaticMode() {
+        this.shallowSomaticMode = true;
```
pauldwolfe, line 73: perhaps we should enforce that somatic mode is enabled as well? otherwise you might end up calling this method, thinking your in shallow and somatic mode.

kduyvesteyn: Added state checking in the build function plus tests for the more plausible errors

cluster/src/test/java/com/hartwig/pipeline/calling/sage/SageCommandBuilderTest.java
```diff
@@ -38,11 +40,25 @@ public void testGermlineCram() {
...
         assertEquals(bash.get(1), "/opt/tools/samtools/1.10/samtools index /data/input/COLO829v003T.bam");
-        assertEquals(bash.get(2), "/opt/tools/samtools/1.10/samtools view -T /opt/resources/reference_genome/37/Homo_sapiens.GRCh37.GATK.illumina.fasta -L /opt/resources/sage/37/SlicePanel.germline.37.bed.gz -o /data/input/COLO829v003R.bam -u -@ $(grep -c '^processor' /proc/cpuinfo) -M /data/input/COLO829v003R.cram");
+        assertEquals(bash.get(2),
+                "/opt/tools/samtools/1.10/samtools view -T /opt/resources/reference_genome/37/Homo_sapiens.GRCh37.GATK.illumina.fasta -L /opt/resources/sage/37/SlicePanel.germline.37.bed.gz -o /data/input/COLO829v003R.bam -u -@ $(grep -c '^processor' /proc/cpuinfo) -M /data/input/COLO829v003R.cram");
         assertEquals(bash.get(3), "/opt/tools/samtools/1.10/samtools index /data/input/COLO829v003R.bam");
         assertEquals(bash.get(4), REFERENCE_SAGE_COMMAND);
     }
 
+    @Test(expected = IllegalStateException.class)
```
pauldwolfe, line 52: Convention is not to prefix tests with `test`... do you mind removing these and also the prefixes from the existing tests?

cluster/src/test/java/com/hartwig/pipeline/calling/sage/SageCommandBuilderTest.java
```diff
@@ -38,11 +40,25 @@ public void testGermlineCram() {
...
+        new SageCommandBuilder(TestInputs.REF_GENOME_37_RESOURCE_FILES).build(Strings.EMPTY);
+    }
+
+    @Test(expected = IllegalStateException.class)
+    public void testExceptionOnShallowInGermline() {
+        new SageCommandBuilder(TestInputs.REF_GENOME_37_RESOURCE_FILES).germlineMode(REFERENCE, REFERENCE_BAM, TUMOR, TUMOR_BAM)
+                .addShallowSomaticMode()
+                .build(Strings.EMPTY);
+    }
+
```
pauldwolfe, line 63: could we add tests for those 2 other failure modes? (somatic + germline and no somatic + no germline)?

pauldwolfe (changes requested):

cluster/src/test/java/com/hartwig/pipeline/calling/sage/SageCommandBuilderTest.java
kduyvesteyn: Renamed

cluster/src/test/java/com/hartwig/pipeline/calling/sage/SageCommandBuilderTest.java
kduyvesteyn: Refactored

PR #293, opened 2021-05-18
DEV-2019 Switch to events 1.2

nedleitch:


cluster/src/main/java/com/hartwig/pipeline/metadata/SomaticMetadataApiProvider.java
```diff
@@ -71,12 +66,7 @@ public SomaticMetadataApi clinicalRun(final Integer setId) {
         return new ClinicalSomaticMetadataApi(run,
                 api.runs(),
                 api.samples(),
-                new StagedOutputPublisher(api.sets(),
-                        sourceBucket,
-                        publisher,
-                        objectMapper,
-                        run,
-                        PipelineStaged.OutputTarget.PATIENT_REPORT),
+                new StagedOutputPublisher(api.sets(), sourceBucket, publisher, objectMapper, run, Context.DIAGNOSTIC),
```
pauldwolfe, line 69: This will mean we can't use the SERVICES or EXPERIMENTAL contexts. I think we need an argument which can be set from the launcher

pauldwolfe (changes requested):

PR #296, opened 2021-05-20
DEV-2024 Use namespaces over filenames to differentiate analysis types

pauldwolfe:
And add the new namespaces for the secondary analysis. The idea being that these are extensions of the
secondary analysis (metrics and metadata).


nedleitch (approved):
I learned something about Predicates reading this. Looks good!

PR #300, opened 2021-06-02
DEV-2048 Use the barcode from the AddDatatype object

pauldwolfe:
If its not present, we leave the barcode empty (we only need barcode when mapping samples with datatype).


cluster/src/main/java/com/hartwig/pipeline/transfer/staged/StagedOutputPublisher.java
```diff
@@ -114,6 +92,17 @@ public void publish(final PipelineState state, final SomaticRunMetadata metadata
         }
     }
 
+    private static PipelineOutputBlob createBlob(final SomaticRunMetadata metadata, final String sampleName,
+            @SuppressWarnings("OptionalUsedAsFieldOrParameterType") final Optional<AddDatatype> dataType, final Blob blobWithMd5) {
```
nedleitch, line 96: Nice suppression.

nedleitch (approved):

PR #302, opened 2021-06-21
Upgraded Linx to 1.16 and configured to plot reportable visualisations

charlesshale:


pauldwolfe (approved):
Although see my other comment about the ticket number in the commit message

PR #303, opened 2021-06-21
Dev 2088

charlesshale:


cluster/src/main/java/com/hartwig/pipeline/tertiary/linx/Linx.java
```diff
@@ -68,6 +73,13 @@ public String namespace() {
                 resourceFiles.ensemblDataCache(),
                 resourceFiles.knownFusionData(),
                 resourceFiles.driverGenePanel()));
+
+        commands.add(new LinxVisualisationsCommand(
```
pauldwolfe, line 77: how is this related to Purple 3.1?

pauldwolfe (changes requested):
The ticket in subject is DEV-2088 which is about upgrading purple not linx. Can we split this into the 2 upgrades?

Also our convention is to add the ticket number to heading of the commit. It's so we can annotate any line of code and quickly go back to the ticket which is responsible for the change.

pauldwolfe:
> The ticket in subject is DEV-2088 which is about upgrading purple not linx. Can we split this into the 2 upgrades?
> 
> Also our convention is to add the ticket number to heading of the commit. It's so we can annotate any line of code and quickly go back to the ticket which is responsible for the change.

pauldwolfe:
Actually I see that the linx stuff is also in another PR, so I guess this was just a GIT issue. If you can remove the overlap we're all good

PR #308, opened 2021-06-25
DEV-2054 Add VirusInterpreter to Pipeline

pauldwolfe:


cluster/src/main/java/com/hartwig/pipeline/datatypes/DataType.java
```diff
@@ -23,6 +23,7 @@
     AMBER,
     VIRUSBREAKEND_SUMMARY,
     VIRUSBREAKEND_VARIANTS,
+    VIRUS_INTERPRETER,
```
pauldwolfe, line 26: (picky one) usually the datatype names describe the resulting output and not the process (variants, summaries, catalogs). Probably VIRUS_INTERPRETATION would be a more consistent name.

cluster/src/main/java/com/hartwig/pipeline/datatypes/FileTypes.java
```diff
@@ -11,6 +11,8 @@
     public static final String CRAI = ".crai";
     public static final String CRAM = ".cram";
 
+    public static final String TSV = "tsv";
```
pauldwolfe, line 14: No leading '.' required?

cluster/src/main/java/com/hartwig/pipeline/tertiary/virus/VirusAnalysis.java
```diff
@@ -10,27 +10,29 @@
...
 import com.hartwig.pipeline.storage.RuntimeBucket;
 import com.hartwig.pipeline.tertiary.TertiaryStage;
-import com.hartwig.pipeline.tools.Versions;
 
-public class VirusBreakend extends TertiaryStage<VirusBreakendOutput> {
+public class VirusAnalysis extends TertiaryStage<VirusOutput> {
 
     public static final String NAMESPACE = "virusbreakend";
 
+    public static final String VIRUS_BREAKEND_VCF = ".virusbreakend.vcf";
```
pauldwolfe, line 29: public?

cluster/src/main/java/com/hartwig/pipeline/tertiary/virus/VirusAnalysis.java
```diff
@@ -10,27 +10,29 @@
...
+    public static final String VIRUS_BREAKEND_SUMMARY = ".virusbreakend.vcf.summary.tsv";
```
pauldwolfe, line 30: There is a way we can remove this duplication between the constructor of the VirusInterpreter substage and this constant by storing the results in a field. See unfilteredVcf in StructuralCaller for an example.

Downside is we end up with a non-final field... I'll leave it to you which evil you want to trade off :)

cluster/src/main/java/com/hartwig/pipeline/datatypes/FileTypes.java
kduyvesteyn: Nope because the extension is not used an "extension upon extension" such as TBI, BAI etc.

Actually ".cram" is the inconsistent one in this list! Will correct it

cluster/src/main/java/com/hartwig/pipeline/tertiary/virus/VirusAnalysis.java
kduyvesteyn: Will leave as-is. I think a bit more dodgy part of this code is that for virus interpreter I specify an output dir and pipeline has to assume a specific file is written. Would have been cleaner to explicitly write to the file the pipeline needs. Will also leave as-is, don't think this will change a lot anyway or is specifically brittle

PR #312, opened 2021-06-30
DEV-2121 Ensure all germline data ends up the germline analysis bucket

pauldwolfe:
Add the sage germline namespace and then look for all files with germline in the name. Bit of a dirty trick
but at this point we don't have another option unless we inventory every file we archive.

### DIFF
--- a/cluster/src/test/java/com/hartwig/pipeline/transfer/staged/StagedOutputPublisherTest.java
+++ b/cluster/src/test/java/com/hartwig/pipeline/transfer/staged/StagedOutputPublisherTest.java
@@ -107,8 +107,23 @@ public class StagedOutputPublisherTest {
     }
 
     @Test
-    public void publishesGermlineTertiaryAnalysisOnGermlineVcf() throws Exception {
-        verifyGermline(".germline.vcf.gz", "reference.germline.vcf.gz");
+    public void publishesGermlineAnalysisOnGatkGermlineVcf() throws Exception {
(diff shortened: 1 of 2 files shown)


cluster/src/test/java/com/hartwig/pipeline/transfer/staged/StagedOutputPublisherTest.java
```diff
@@ -107,8 +107,23 @@ public void publishesDnaTertiaryAnalysisOnVcf() throws Exception {
...
-    public void publishesGermlineTertiaryAnalysisOnGermlineVcf() throws Exception {
-        verifyGermline(".germline.vcf.gz", "reference.germline.vcf.gz");
+    public void publishesGermlineAnalysisOnGatkGermlineVcf() throws Exception {
+        verifyGermline(".germline.vcf.gz", "reference.germline.vcf.gz", "/germline_caller/");
+    }
+
+    @Test
+    public void publishesGermlineAnalysisOnGatkGermlineVcfIndex() throws Exception {
+        verifyGermline(".germline.vcf.gz.tbi", "reference.germline.vcf.gz.tbi", "/germline_caller/");
+    }
```
nedleitch, line 117: Do we need to add a test for a file that's in a germline namespace but which doesn't have the string "germline" in its name? Does any such file exist?

nedleitch (approved):

PR #316, opened 2021-08-11
DEV-2186 Populate DOIDs in research metadata api

pauldwolfe:
The research api implementation is used in reruns, to this point didn't populate the DOID on the
tumor sample.


stefvanlieshout:
Looks logical to me but a bit out of my comfort zone so for what's it worth!

nedleitch (approved):

PR #331, opened 2021-10-15
DEV-2227 Populate start and end time on both diag and research runs

pauldwolfe:
Diagnostics already populated the status, so extracted that code to be shared between the apis and added
the time population to them.  Using a standard ISO format for date (UTC)

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/metadata/ApiRunStatus.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/metadata/ApiRunStatus.java
@@ -4,32 +4,20 @@ import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.hartwig.api.RunApi;
 import com.hartwig.api.model.Run;
 import com.hartwig.api.model.RunFailure;
 import com.hartwig.api.model.Status;
 import com.hartwig.api.model.UpdateRun;
 import com.hartwig.pipeline.execution.PipelineStatus;
-import com.hartwig.pipeline.jackson.ObjectMappers;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class ApiRunStatus {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(ApiRunStatus.class);
     private static final String PIPELINE_SOURCE = "Pipeline";
     private static final String HEALTH_CHECK = "HealthCheck";
 
     static void start(final RunApi runApi, final Run run) {
-        UpdateRun update = new UpdateRun().failure(null).status(Status.PROCESSING).startTime(timestamp());
-        try {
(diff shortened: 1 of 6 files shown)


cluster/src/test/java/com/hartwig/pipeline/metadata/ResearchMetadataApiTest.java
```diff
@@ -208,6 +220,32 @@ public void doesNothingOnFailedPipeline() {
...
 
+    @Test
+    public void setsStatusAndStartTimeOnStart() {
+        ArgumentCaptor<Long> runIdArgumentCaptor = ArgumentCaptor.forClass(Long.class);
+        ArgumentCaptor<UpdateRun> updateRunArgumentCaptor = ArgumentCaptor.forClass(UpdateRun.class);
+        when(runApi.update(runIdArgumentCaptor.capture(), updateRunArgumentCaptor.capture())).thenReturn(run);
+        victim.start();
+        assertThat(runIdArgumentCaptor.getValue()).isEqualTo(RUN_ID);
+        UpdateRun updateRun = updateRunArgumentCaptor.getValue();
+        assertThat(updateRun.getStartTime()).isNotNull();
```
nedleitch, line 231: Would the AssertJ's `isCloseTo` be appropriate here? Would definitely be more rigorous than just making sure it's non-null.

cluster/src/main/java/com/hartwig/pipeline/metadata/ApiRunStatus.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(ApiRunStatus.class);
+    private static final String PIPELINE_SOURCE = "Pipeline";
+    private static final String HEALTH_CHECK = "HealthCheck";
+
+    static void start(final RunApi runApi, final Run run) {
+        UpdateRun update = new UpdateRun().failure(null).status(Status.PROCESSING).startTime(timestamp());
+        try {
+            LOGGER.info("Updating API with run status and result [{}]", ObjectMappers.get().writeValueAsString(update));
+        } catch (JsonProcessingException e) {
+            e.printStackTrace();
```
nedleitch, line 29: This is a bit of a funny one, maybe we should write a log message explaining the stacktrace too?

cluster/src/main/java/com/hartwig/pipeline/metadata/ApiRunStatus.java
```diff
@@ -0,0 +1,54 @@
...
+            case QC_FAILED:
+                return new UpdateRun().status(Status.FAILED)
+                        .failure(new RunFailure().type(RunFailure.TypeEnum.QCFAILURE).source(HEALTH_CHECK));
+            default:
+                return new UpdateRun().status(Status.FINISHED);
+        }
+    }
+
+    private static String timestamp() {
+        return LocalDateTime.now().format(DateTimeFormatter.ISO_DATE_TIME);
```
nedleitch, line 52: Is it worth making the timezone explicitly UTC rather than relying on the server to have its timezone set correctly?

nedleitch (changes requested):

PR #341, opened 2021-11-10
DEV-2347 Integrating the new events API

pauldwolfe:
Combines the 3 events into a single event composed of 3 analyses.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/Arguments.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/Arguments.java
@@ -97,7 +97,7 @@ public interface Arguments extends CommonArguments {
 
     boolean anonymize();
 
-    Pipeline.Context analysisContext();
+    Pipeline.Context context();
 
     static String workingDir() {
         return System.getProperty("user.dir");
@@ -123,7 +123,7 @@ public interface Arguments extends CommonArguments {
 
     int DEFAULT_MAX_CONCURRENT_LANES = 8;
 
-    Pipeline.Context DEFAULT_ANALYSIS_CONTEXT = Context.DIAGNOSTIC;
+    Pipeline.Context DEFAULT_CONTEXT = Context.DIAGNOSTIC;
 
     static ImmutableArguments.Builder defaultsBuilder(String profileString) {
         DefaultsProfile profile = DefaultsProfile.valueOf(profileString.toUpperCase());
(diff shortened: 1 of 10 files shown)


cluster/src/main/java/com/hartwig/pipeline/Arguments.java
```diff
@@ -97,7 +97,7 @@ static Arguments testDefaults() {
 
     boolean anonymize();
 
-    Analysis.Context analysisContext();
+    Pipeline.Context analysisContext();
```
nedleitch, line 100: Do we want to remove "analysis" from the names?

cluster/src/main/java/com/hartwig/pipeline/CommandLineOptions.java
```diff
@@ -373,7 +374,7 @@ public static Arguments from(String[] args) throws ParseException {
 
     public static Context analysisContext(final CommandLine commandLine, final Arguments defaults) {
```
nedleitch, line 375: Does seem like we're being inconsistent here, is it still an analysis context (ie pipeline analysis context)? Or is it a pipeline context?

cluster/src/main/java/com/hartwig/pipeline/transfer/staged/StagedOutputPublisher.java
```diff
@@ -69,30 +72,38 @@ public void publish(final PipelineState state, final SomaticRunMetadata metadata
...
-            publish(tertiaryAnalysisEvent.build());
-            publish(germlineAnalysisEvent.build());
+            publish(ImmutablePipelineStaged.builder()
+                    .pipeline(ImmutablePipeline.builder()
+                            .sample(sampleName)
+                            .bucket(sourceBucket.getName())
+                            .runId(run.getId())
+                            .setId(set.getId())
+                            .context(context)
+                            .addAnalyses(alignedReadsAnalysis.build(), somaticAnalysis.build(), germlineAnalysis.build())
```
nedleitch, line 99: This is so much nicer!

cluster/src/test/java/com/hartwig/pipeline/metadata/ResearchMetadataApiTest.java
```diff
@@ -151,12 +153,14 @@ public void publishesPipelineStagedEventOnCompletionSomaticFile() throws Excepti
...
-        assertThat(result.setId()).isEqualTo(SET_ID);
-        assertThat(result.sample()).isEqualTo("tumor");
-        assertThat(result.version()).isEqualTo(Versions.pipelineMajorMinorVersion());
-        PipelineOutputBlob blobResult = result.blobs().get(0);
+        assertThat(result.pipeline().runId()).isEqualTo(1);
+        assertThat(result.pipeline().setId()).isEqualTo(SET_ID);
+        assertThat(result.pipeline().sample()).isEqualTo("tumor");
+        assertThat(result.pipeline().version()).isEqualTo(Versions.pipelineMajorMinorVersion());
+
+        Analysis analysis = result.pipeline().analyses().get(1);
```
nedleitch, line 161: Think the testcase names need some attention so they align with what you're doing here (picking out one part of the single event that is published).

nedleitch (changes requested):

nedleitch (approved):

PR #342, opened 2021-11-18
DEV-2358 Orange upgrade to 1.4

pauldwolfe:
Version bump plus a couple new resources.

Also implemented a fix here to check for the existence in GCS of the feature plot. This enables use
of the Orange 1.2 feature to pass a non-existent cuppa feature plot as an argument (the feature plot
is not always used).

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/execution/vm/InputDownloadIfBlobExists.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/execution/vm/InputDownloadIfBlobExists.java
@@ -1,0 +1,26 @@
+package com.hartwig.pipeline.execution.vm;
+
+import com.hartwig.pipeline.storage.GoogleStorageLocation;
+
+public class InputDownloadIfBlobExists implements BashCommand {
+
+    private final InputDownload download;
+    private final GoogleStorageLocation googleStorageLocation;
+
+    public InputDownloadIfBlobExists(final GoogleStorageLocation googleStorageLocation) {
+        this.download = new InputDownload(googleStorageLocation);
+        this.googleStorageLocation = googleStorageLocation;
+    }
+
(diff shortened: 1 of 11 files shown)


cluster/src/main/java/com/hartwig/pipeline/tertiary/orange/Orange.java
```diff
@@ -245,7 +243,12 @@ public String namespace() {
                                 "-annotated_virus_tsv",
                                 annotatedVirusTsv.getLocalTargetPath(),
                                 "-pipeline_version_file",
-                                pipelineVersionFilePath)));
+                                pipelineVersionFilePath,
+                                "",
```
nedleitch, line 247: Think this can be removed.

cluster/src/main/java/com/hartwig/pipeline/execution/vm/InputDownloadIfBlobExists.java
```diff
@@ -0,0 +1,26 @@
...
+    private final GoogleStorageLocation googleStorageLocation;
+
+    public InputDownloadIfBlobExists(final GoogleStorageLocation googleStorageLocation) {
+        this.download = new InputDownload(googleStorageLocation);
+        this.googleStorageLocation = googleStorageLocation;
+    }
+
+    @Override
+    public String asBash() {
+        return String.format("gsutil  -q stat  gs://%s/%s; if [ $? == 0 ]; then  %s ; fi",
```
nedleitch, line 17: Could be shortened to `gsutil -q stat gs://%s/%s && %s` I believe if brevity is important; that's a pretty well-known idiom so I don't think readability would suffer either.

nedleitch (changes requested):

PR #353, opened 2021-12-31
DEV-2464 Add -K to BWA mem command

pauldwolfe:


cluster/src/main/java/com/hartwig/pipeline/alignment/bwa/BwaMemCommand.java
```diff
@@ -15,6 +15,8 @@
                 "-Y",
                 "-t",
                 Bash.allCpus(),
+                "-K",
+                "10000000",
```
nedleitch, line 19: Could you pull this into a constant just to try to clarify why it's here?

nedleitch (changes requested):

cluster/src/main/java/com/hartwig/pipeline/alignment/bwa/BwaMemCommand.java
pauldwolfe: Done

PR #355, opened 2022-01-17
DEV-2492 Add new labels and labelling for disks and buckets

pauldwolfe:
Introduces a couple new labels: cost_center, which can be used to bucket any cost;
user, the instigating user which will be defaulted when run locally; sample, the
tumor sample name or ref if tumor not available. Run ID also changes, removing
superfuous information from the RunTag (so run_id argument is used exactly).

Labels are refactored to be used both in the compute and storage contexts and
also in the batch framework consistently.


cluster/src/test/java/com/hartwig/pipeline/storage/RuntimeBucketTest.java
```diff
@@ -137,12 +144,30 @@ public void namespacesListOperationPrefix() {
...
+                arguments,
+                Labels.of(arguments, TestInputs.defaultSomaticRunMetadata()));
+        assertThat(bucketInfo.getValue().getLabels()).isEqualTo(Map.of("cost_center",
+                "development",
+                "run_id",
+                "test",
+                "sample",
+                "tumor",
+                "user",
+                "pwolfe"));
```
nedleitch, line 166: See above comments on `GoogleComputeEngineTest`.

cluster/src/main/java/com/hartwig/pipeline/labels/Labels.java
```diff
@@ -1,19 +1,58 @@
...
+        runId().ifPresent(l -> mapBuilder.put("run_id", clean(l)));
+        user().ifPresent(l -> mapBuilder.put("user", clean(l)));
+        costCenter().ifPresent(l -> mapBuilder.put("cost_center", clean(l)));
+        for (Map.Entry<String, String> entry : additional) {
+            mapBuilder.put(entry);
+        }
+        return mapBuilder.build();
     }
 
     private static String clean(final String run) {
```
nedleitch, line 58: Nit: rename parameter?

nedleitch (changes requested):

PR #363, opened 2022-02-11
DEV-1933 Add Lilac to pipeline

nedleitch:
For the moment the output is not used by any other tools.

Also updated the imaging script to allow multiple instances concurrently,
with the local username discriminating them.


cluster/images/create_public_image.sh
```diff
@@ -13,16 +13,17 @@ while getopts ':tf:' flag; do
         *) ;;
     esac
 done
-sourceInstance="pipeline5-${VERSION}${FLAVOUR+"-$FLAVOUR"}"
-imageName="${sourceInstance}-$(date +%Y%m%d%H%M)"
-image_project="debian-cloud"
-image_family="debian-9"
+image_family="pipeline5-${VERSION}${FLAVOUR+"-$FLAVOUR"}"
```
pauldwolfe, line 16: maybe unrelated?

cluster/src/main/java/com/hartwig/pipeline/tertiary/lilac/Lilac.java
```diff
@@ -0,0 +1,105 @@
...
+                .build();
+    }
+
+    @Override
+    public LilacOutput output(final SomaticRunMetadata metadata, final PipelineStatus jobStatus, final RuntimeBucket bucket,
+            final ResultsDirectory resultsDirectory) {
+        return LilacOutput.builder()
+                .status(jobStatus)
+                .addFailedLogLocations(GoogleStorageLocation.of(bucket.name(), RunLogComponent.LOG_FILE))
+                .addReportComponents(new EntireOutputComponent(bucket, Folder.root(), namespace(), resultsDirectory))
```
pauldwolfe, line 92: We'll want to add the Lilac loader to the DB loaders so we should add a datatype.

cluster/src/main/java/com/hartwig/pipeline/tertiary/lilac/LilacBamSliceCommand.java
```diff
@@ -0,0 +1,22 @@
+package com.hartwig.pipeline.tertiary.lilac;
+
+import com.hartwig.pipeline.execution.vm.BashCommand;
+import com.hartwig.pipeline.execution.vm.SambambaCommand;
+import com.hartwig.pipeline.resource.ResourceFiles;
+
+public class LilacBamSliceCommand implements BashCommand {
```
pauldwolfe, line 7: its a little nicer to extend SambambaCommand rather than semi-decorate it.

pauldwolfe (changes requested):

pauldwolfe (approved):

PR #365, opened 2022-02-16
DEV-2547 Improvements to imager scripts

nedleitch:


stefvanlieshout (approved):
Nice solution to allow testing of pilot versions 👍

PR #368, opened 2022-02-28
DEV-2573 Add panel bed to common resources

pauldwolfe:
The panel bed is configured as an optional arg and can be either a location within the
image or a location in google cloud storage

The bed is integrated with the resource files in a similar manner to the reference genome
where the GCS based bed is auto-downloaded by the stage runner prior to stage start.

Stages can then decide whether to use the panel bed via ResourceFiles

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/CommonArguments.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/CommonArguments.java
@@ -77,6 +77,8 @@ public interface CommonArguments {
 
     Optional<String> userLabel();
 
+    Optional<String> panelBedLocation();
+
     static Optional<String> privateKey(CommandLine commandLine) {
         if (commandLine.hasOption(PRIVATE_KEY_PATH)) {
             return Optional.of(commandLine.getOptionValue(PRIVATE_KEY_PATH));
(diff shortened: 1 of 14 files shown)


cluster/src/main/java/com/hartwig/pipeline/CommonArguments.java
```diff
@@ -77,6 +77,8 @@
 
     Optional<String> userLabel();
 
+    Optional<String> panelBedLocation();
```
nedleitch, line 80: Do you foresee this being used in the batch framework too? Or would it be better in the pipeline-specific arguments?

cluster/src/main/java/com/hartwig/pipeline/tertiary/HmfToolCommandBuilder.java
```diff
@@ -36,6 +39,12 @@ public HmfToolCommandBuilder reference(final String referenceSample, final Strin
         return this;
     }
 
+    @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+    public HmfToolCommandBuilder panelBed(final Optional<String> panelBedLocation) {
+      //  arguments.addAll(panelBedLocation.stream().flatMap(l -> Stream.of("-panel_bed", l)).collect(Collectors.toList()));
```
nedleitch, line 44: This is commented just because not all commands support it yet?

nedleitch (approved):

cluster/src/main/java/com/hartwig/pipeline/CommonArguments.java
pauldwolfe: I do foresee. The guys and aus will be doing a bunch of testing with panel

cluster/src/main/java/com/hartwig/pipeline/tertiary/HmfToolCommandBuilder.java
pauldwolfe: yes

PR #374, opened 2022-03-04
DEV-2564 Publish event for verification

nedleitch:
A second attempt at this, with fewer changes in-pipeline. The loader will bear
the brunt of the changes.

With this approach a new command-line argument has been added which will govern
the publishing of the `pipeline.completed` event when in "local" metadata mode.
If this argument is given the StagedOutputPublisher will be invoked to publish
the event; it is the same event that gets published in the regular RESEARCH and
DIAGNOSTIC contexts when an actual API is involved.

The `genomic-db-loader` is being changed to directly load the event which is
usually processed by the `archiver`. We will assume that pipelines will only be
run with this flag in an environment where it is appropriate, to avoid having
the `archiver` intercept these events where not desired.


cluster/src/main/java/com/hartwig/pipeline/Arguments.java
```diff
@@ -125,7 +128,7 @@ static String workingDir() {
 
     Pipeline.Context DEFAULT_CONTEXT = Context.DIAGNOSTIC;
 
-    static ImmutableArguments.Builder defaultsBuilder(String profileString) {
```
pauldwolfe, line 128: small nit, but i prefer not to on-demand static import this, otherwise can be tricky to see right away what type of builder it is.

cluster/src/main/java/com/hartwig/pipeline/Arguments.java
```diff
@@ -158,6 +161,7 @@ static String workingDir() {
                     .network(DEFAULT_NETWORK)
                     .outputCram(true)
                     .publishToTurquoise(false)
+                    .publishLoadEvent(false)
```
pauldwolfe, line 164: name is a bit cryptic

cluster/src/test/java/com/hartwig/pipeline/metadata/LocalSomaticMetadataTest.java
```diff
@@ -3,19 +3,26 @@
...
 import com.hartwig.pipeline.Arguments;
+import com.hartwig.pipeline.PipelineState;
 import com.hartwig.pipeline.alignment.sample.JsonSampleSource;
 import com.hartwig.pipeline.metadata.SingleSampleRunMetadata.SampleType;
+import com.hartwig.pipeline.transfer.staged.StagedOutputPublisher;
 
 import org.junit.Before;
 import org.junit.Test;
 
+import net.bytebuddy.build.ToStringPlugin.Enhance;
```
pauldwolfe, line 24: unused?

cluster/src/main/java/com/hartwig/pipeline/metadata/LocalSomaticMetadata.java
```diff
@@ -8,15 +8,18 @@
...
+import com.hartwig.pipeline.transfer.staged.StagedOutputPublisher;
 
 public class LocalSomaticMetadata implements SomaticMetadataApi {
 
     private final Arguments arguments;
     private final JsonSampleSource jsonSampleSource;
+    private final StagedOutputPublisher stagedOutputPublisher;
 
-    LocalSomaticMetadata(final Arguments arguments, JsonSampleSource jsonSampleSource) {
+    LocalSomaticMetadata(final Arguments arguments, JsonSampleSource jsonSampleSource, StagedOutputPublisher stagedOutputPublisher) {
```
pauldwolfe, line 19: final

cluster/src/main/java/com/hartwig/pipeline/metadata/SomaticMetadataApiProvider.java
```diff
@@ -44,47 +45,45 @@ public SomaticMetadataApi get() {
...
-                        objectMapper,
+                createPublisher(SetResolver.forApi(api.sets()),
                         Optional.of(run),
                         arguments.context(),
-                        arguments.outputCram(),
                         false),
                 new Anonymizer(arguments));
     }
+
+    private StagedOutputPublisher createPublisher(SetResolver setResolver, Optional<Run> run, Context context, boolean useOnlyDBSets) {
```
pauldwolfe, line 84: final

pauldwolfe (changes requested):

PR #376, opened 2022-03-08
DEV-2596 Refactor gripss somatic/germline to consolidate shared logic

pauldwolfe:
Follows the SAGE structure moving the differences between somatic and germline out into
a configuration class. 

In preparation for the germline/tumor only changes we create a new builder for the gripss
command. Cobalt and Amber commands are changed to JAR commands, turns out they always had
an entry point defined.

As of this change tumor-reference is still the only mode enabled as GRIDSS is still 
disabled in both modes.


cluster/src/main/java/com/hartwig/pipeline/tertiary/cobalt/CobaltCommandBuilder.java
```diff
@@ -12,7 +12,7 @@
     private static final String COBALT = "cobalt";
 
     public CobaltCommandBuilder(final ResourceFiles resourceFiles) {
-        super(COBALT, Versions.COBALT, MAX_HEAP, MAIN_CLASS, JAR);
+        super(COBALT, Versions.COBALT, MAX_HEAP, JAR);
```
nedleitch, line 15: Can the `MAIN_CLASS` constant be removed too then?

cluster/src/main/java/com/hartwig/pipeline/tertiary/amber/AmberCommandBuilder.java
```diff
@@ -11,7 +11,7 @@
     private static final String MAIN_CLASS = "com.hartwig.hmftools.amber.AmberApplication";
 
     private AmberCommandBuilder(final ResourceFiles resourceFiles) {
-        super("amber", Versions.AMBER, MAX_HEAP, MAIN_CLASS, JAR);
+        super("amber", Versions.AMBER, MAX_HEAP, JAR);
```
nedleitch, line 14: Can we remove `MAIN_CLASS`? Also I question know what the other constants add when they're used only once and their contents are pretty self-explanatory.

cluster/src/main/java/com/hartwig/pipeline/tertiary/HmfToolCommandBuilder.java
```diff
@@ -10,21 +10,20 @@
 import com.hartwig.pipeline.execution.vm.BashCommand;
 import com.hartwig.pipeline.execution.vm.VmDirectories;
 import com.hartwig.pipeline.execution.vm.java.JavaClassCommand;
```
nedleitch, line 12: Unused?

nedleitch (changes requested):

PR #378, opened 2022-03-08
DEV-2599 Port GRIDSS to new input modes

pauldwolfe:
Biggest change is to the Driver class, which now takes the tumor and reference sample names and
paths in a builder style.

The driver is configured to pass different permutations of input bams (last argument) based on which
mode we're running.

Renamed the stage and output to Gridss and Gridss to move towards the more explicit naming found
elsewhere in the pipeline.


cluster/src/main/java/com/hartwig/pipeline/calling/structural/gridss/stage/Driver.java
```diff
@@ -12,52 +15,79 @@
...
+            return sampleName;
+        }
+
+        public String getSamplePath() {
+            return samplePath;
+        }
+
+        @Override
+        public int compareTo(@NotNull final Driver.SampleArgument o) {
+            return this.type.compareTo(o.type);
```
nedleitch, line 90: Am I understanding properly that this just maintains the order of reference first if it exists?

cluster/src/main/java/com/hartwig/pipeline/calling/structural/gridss/Gridss.java
```diff
@@ -35,20 +33,21 @@
...
 import com.hartwig.pipeline.reruns.PersistedLocations;
 import com.hartwig.pipeline.resource.ResourceFiles;
-import com.hartwig.pipeline.stages.Stage;
 import com.hartwig.pipeline.stages.SubStageInputOutput;
 import com.hartwig.pipeline.storage.GoogleStorageLocation;
 import com.hartwig.pipeline.storage.RuntimeBucket;
 import com.hartwig.pipeline.tertiary.TertiaryStage;
 
-public class StructuralCaller extends TertiaryStage<StructuralCallerOutput> {
+import org.jetbrains.annotations.NotNull;
```
nedleitch, line 41: Not trying to be controversial, what is the benefit of adding this annotation beyond dev-time static analysis from within Intellij?

nedleitch (approved):

cluster/src/main/java/com/hartwig/pipeline/calling/structural/gridss/stage/Driver.java
pauldwolfe: yeah, i'll make it a bit more clear

cluster/src/main/java/com/hartwig/pipeline/calling/structural/gridss/Gridss.java
pauldwolfe: accidental addition, i'll remove.

PR #388, opened 2022-03-25
DEV-2634 Disable "release" (aka docker) profile by default

pauldwolfe:
The release profile no longer creates a docker container, but still creates the runnable
jar and copies the dependencies to the correct locations for the containerization. So we
call it release and not "smoke test".

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,11 +75,6 @@
                 <version>${java-client.version}</version>
             </dependency>
             <dependency>
-                <groupId>com.hartwig</groupId>
-                <artifactId>compar</artifactId>
-                <version>${compar.version}</version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -209,6 +204,12 @@
                 <version>${mockito.version}</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
+                <groupId>com.hartwig</groupId>
+                <artifactId>compar</artifactId>
+                <version>${compar.version}</version>
+                <scope>test</scope>
```
nedleitch, line 211: Nice catch, thanks for adding this!

nedleitch (approved):

PR #391, opened 2022-03-30
DEV-2593 Support tumor-only, germline-only and target-region modes

pauldwolfe:
Refer to ticket DEV-2593 and supporting doc for details


cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageCaller.java
```diff
@@ -35,8 +35,8 @@
     public static final String SAGE_GENE_COVERAGE_TSV = "sage.gene.coverage.tsv";
     public static final String SAGE_BQR_PNG = "sage.bqr.png";
 
-    private final PersistedDataset persistedDataset;
-    private final SageConfiguration sageConfiguration;
+    protected final PersistedDataset persistedDataset;
```
pauldwolfe, line 38: These protected fields are not accessed by the subclasses, so I think they can remain private

cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageGermlinePostProcess.java
```diff
@@ -13,25 +13,25 @@
...
     @Override
     public List<BashCommand> bash(final OutputFile input, final OutputFile output) {
         final List<BashCommand> result = Lists.newArrayList();
 
         SubStage passFilter = new PassFilter();
+
+        OutputFile finalOutputFile = OutputFile.of(tumorSampleName.sampleName(), SAGE_GERMLINE_FILTERED, FileTypes.GZIPPED_VCF);
+        result.addAll(passFilter.bash(input, finalOutputFile));
+
+        /*
```
pauldwolfe, line 32: can be removed?

cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageSomaticPostProcess.java
```diff
@@ -13,34 +13,34 @@
...
 
-        OutputFile passFilterFile = passFilter.apply(tumorSampleName).outputFile();
-        OutputFile mappabilityAnnotationFile = mappabilityAnnotation.apply(tumorSampleName).outputFile();
-        OutputFile ponAnnotationFile = ponAnnotation.apply(tumorSampleName).outputFile();
-        OutputFile ponFilterFile = ponFilter.apply(tumorSampleName).outputFile();
+        OutputFile finalOutputFile = OutputFile.of(tumorSampleName.sampleName(), SAGE_SOMATIC_FILTERED, FileTypes.GZIPPED_VCF);
+        result.addAll(passFilter.bash(input, finalOutputFile));
 
-        result.addAll(passFilter.bash(input, passFilterFile));
+        /*
```
pauldwolfe, line 32: Can be removed?

cluster/src/main/java/com/hartwig/pipeline/calling/structural/gripss/Gripss.java
```diff
@@ -157,4 +154,10 @@ public GripssOutput persistedOutput(final SomaticRunMetadata metadata) {
     public boolean shouldRun(final Arguments arguments) {
         return arguments.runTertiary();
     }
+
+    public String filteredVcf(final SomaticRunMetadata metadata) { return ""; }
```
pauldwolfe, line 158: These should be abstract rather than returning empty strings and null

cluster/src/main/java/com/hartwig/pipeline/calling/structural/gripss/GripssGermline.java
```diff
@@ -11,13 +17,57 @@
...
 public class GripssGermline extends Gripss {
 
+    public static final String GERMLINE_NAMESPACE = "gripss_germline";
+
+    private static final String GRIPSS_GERMLINE_FILTERED = ".gripss.filtered.germline.";
+    private static final String GRIPSS_GERMLINE_UNFILTERED = ".gripss.germline.";
+
     public GripssGermline(final GridssOutput gridssOutput, final PersistedDataset persistedDataset,
             final ResourceFiles resourceFiles) {
-        super(gridssOutput, persistedDataset, GripssConfiguration.germline(resourceFiles));
```
pauldwolfe, line 16: Not sure why you re-refactored this to use abstract template methods again. I thought we wanted to reduce the code in the subclasses? Which is why I created the GripssConfiguration class to go for a more compositional approach. 

I don't really mind that much one way or the other, but would be good to decide what is the preferred approach and keep it consistent (for instance now Sage uses SageConfiguration over a bunch of template methods).

cluster/src/main/java/com/hartwig/pipeline/resource/RefGenome38ResourceFiles.java
```diff
@@ -186,4 +195,24 @@ public String hlaRegionBed() {
...
     public String purpleCohortGermlineDeletions() { return formPath(PURPLE, "cohort_germline_del_freq.38.csv"); }
 
+    @Override
+    public void setTargetRegionsDir(final String targetRegionsDir) {
+        this.targetRegionsDir = targetRegionsDir;
+    }
+
+    @Override
+    public Optional<String> targetRegionsBed() {
+        if(targetRegionsDir == null)
```
pauldwolfe, line 205: You can also use a Optional.ofNullable().map() to simplify this a little

cluster/src/main/java/com/hartwig/pipeline/resource/ResourceFilesFactory.java
```diff
@@ -9,13 +9,15 @@ public static ResourceFiles buildResourceFiles(final RefGenomeVersion version) {
...
         ResourceFiles resourceFiles = buildResourceFiles(arguments.refGenomeVersion());
+
         if (arguments.refGenomeUrl().isPresent()) {
             resourceFiles = new OverriddenReferenceGenome(resourceFiles, arguments.refGenomeUrl().get());
         }
-        if (arguments.targetRegionsBedLocation().isPresent()) {
-            resourceFiles = new TargetRegionsEnabled(resourceFiles, arguments.targetRegionsBedLocation().get());
-        }
+
+        TargetRegionsFiles.parseArguments(resourceFiles, arguments);
```
pauldwolfe, line 19: From a quick scan it's difficult to see what this is doing. The method is called parseArguments but doesn't return anything, so it must actually be mutating the arguments. Mutating arguments to a function is something we avoid on principle.

cluster/src/main/java/com/hartwig/pipeline/resource/TargetRegionsFiles.java
```diff
@@ -0,0 +1,41 @@
...
+        commands.add(new MkDirCommand(OVERRIDE_TARGETED_REGIONS_DIR));
+
+        String targetRegionsDir = arguments.targetRegionsDir().get();
+        commands.add(() -> String.format("gsutil -qm cp -n %s/* %s", targetRegionsDir, OVERRIDE_TARGETED_REGIONS_DIR));
+        return commands;
+    }
+
+    public static void parseArguments(final ResourceFiles resourceFiles, final CommonArguments arguments) {
+        if (arguments.useTargetRegions()) {
+            resourceFiles.setTargetRegionsDir(TARGETED_REGIONS_DIR);
```
pauldwolfe, line 35: As mentioned we try not to mutate incoming parameters, but also we try to void mutability in general especially on the interfaces abstracting access to data. Now clients have to know something about the order of things, ie they could call getTargetRegionsDir before the setter is called and assume we're not in target mode. This is why I first implemented with a decorator on an immutable resource files. 

I'm ok to leave for now, as its more about future proofing. But will likely come back to this before next release.

cluster/src/main/java/com/hartwig/pipeline/tertiary/amber/Amber.java
```diff
@@ -45,11 +48,58 @@ public String namespace() {
...
     public List<BashCommand> tumorReferenceCommands(final SomaticRunMetadata metadata) {
-        return List.of(AmberCommandBuilder.newBuilder(resourceFiles)
-                .tumor(metadata.tumor().sampleName(), getTumorBamDownload().getLocalTargetPath())
-                .reference(metadata.reference().sampleName(), getReferenceBamDownload().getLocalTargetPath())
-                .targetRegionsBed(resourceFiles.targetRegionsBed())
-                .build(), new CopyResourceToOutput(resourceFiles.amberSnpcheck()));
+
+        List<String> arguments = Lists.newArrayList();
+
+        arguments.add(String.format("-tumor %s", metadata.tumor().sampleName()));
```
pauldwolfe, line 54: I don't see an improvement here by removing the builder. It's a little more straightforward this way, but reduces but introduces a bunch of duplication and reduces encapsulation.

cluster/src/main/java/com/hartwig/pipeline/tertiary/lilac/Lilac.java
```diff
@@ -61,19 +65,85 @@ public String namespace() {
...
-                new LilacBamSliceCommand(resourceFiles, getTumorBamDownload().getLocalTargetPath(), slicedTumorBam),
-                new LilacBamIndexCommand(slicedTumorBam),
-                new LilacCommand(resourceFiles,
-                        metadata.tumor().sampleName(),
-                        slicedRefBam,
-                        slicedTumorBam,
-                        purpleGeneCopyNumber.getLocalTargetPath(),
-                        purpleSomaticVariants.getLocalTargetPath()));
+
+        List<BashCommand> commands = Lists.newArrayList();
```
pauldwolfe, line 73: This also seems like a subjective refactor. It's not really better or worse than the old way, just different. In general we break down the stages in terms of responsibilities:
- The stage wires things together (input, output, commands)
- Individual command classes define the commands

These new stages are now inconsistent with all the rest and will start to create two camps within the codebase.

cluster/src/main/java/com/hartwig/pipeline/tertiary/linx/LinxGermline.java
```diff
@@ -56,16 +63,32 @@ public String namespace() {
...
-                resourceFiles.gridssBreakendPon()));
+    public List<BashCommand> tumorReferenceCommands(final SomaticRunMetadata metadata) { return buildCommand(metadata, TUMOR_REFERENCE); }
+
+    @Override
+    public List<BashCommand> referenceOnlyCommands(final SomaticRunMetadata metadata) { return buildCommand(metadata, REFERENCE_ONLY); }
+
+    @Override
+    public List<BashCommand> tumorOnlyCommands(final SomaticRunMetadata metadata) { return Stage.disabled(); }
+
+    private List<BashCommand> buildCommand(final SomaticRunMetadata metadata, final InputMode inputMode) {
```
pauldwolfe, line 74: we could pass the sample name in rather than the metadata and remove the need to check the mode

cluster/src/main/java/com/hartwig/pipeline/tertiary/pave/PaveArgumentBuilder.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.List;
+
+import com.google.api.client.util.Lists;
+import com.hartwig.pipeline.execution.vm.VmDirectories;
+import com.hartwig.pipeline.metadata.InputMode;
+import com.hartwig.pipeline.resource.RefGenomeVersion;
+import com.hartwig.pipeline.resource.ResourceFiles;
+
+public class PaveArgumentBuilder
```
pauldwolfe, line 13: This class isn't a builder, maybe just call it PaveArguments?

Elsewhere you've inlined builders/command classes and used the class hierarchy to share common code. Would it be possible to inline this class as well for consistency?

cluster/src/main/java/com/hartwig/pipeline/CommandLineOptions.java
```diff
@@ -143,13 +144,18 @@ private static Options options() {
...
 
-    private static Option targetRegionsBedLocation() {
-        return optionWithArg(TARGET_REGIONS_BED_FLAG,
-                "Specify the location of the target regions BED to enable targeted mode. Both within image and gs:// urls are supported.");
+    private static Option targetRegionsDir() {
+        return optionWithArg(TARGET_REGIONS_DIR_FLAG,
+                "Specify the location of the target regions files and enable target-regions mode. Both within image and gs:// urls are supported.");
+    }
+
+    private static Option useTargetRegions() {
```
pauldwolfe, line 157: What's the rationale for adding an additional boolean parameter? Can we not just detect the mode via the presence of the directory?

PR #399, opened 2022-04-11
DEV-2687 Remove option to override or specify target regions

pauldwolfe:
Was premature to add the override behaviour and complicated the resource file
with optionals and setters.

This commit removes all optionality from the resource files instead using the
argument as the central toggle for target regions.


cluster/src/main/java/com/hartwig/pipeline/resource/RefGenome37ResourceFiles.java
```diff
@@ -22,6 +22,8 @@
 
 import java.util.Optional;
 
+import org.jetbrains.annotations.NotNull;
```
nedleitch, line 25: Unused?

nedleitch (approved):

PR #406, opened 2022-05-04
DEV-2705 Add starting point for Protect-only runs

nedleitch:
This can probably use some work as several of the later steps are run, not just
Protect. Also a couple of the recently-added stages were never added to the
existing starting points so I have added those.

I have also added a `Namespace` annotation for the stages and a couple of tests
that use reflection to do some basic checks on the namespaces and starting
points. This may be a first step toward a more cohesive structure for the
starting points but also is a move closer to a more official interface for the
stages. For the moment almost all of the stages reference existing constants
as the value for their new annotation but this too is just a first step.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageCaller.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageCaller.java
@@ -28,7 +28,7 @@ import com.hartwig.pipeline.storage.GoogleStorageLocation;
 import com.hartwig.pipeline.storage.RuntimeBucket;
 import com.hartwig.pipeline.tertiary.TertiaryStage;
 
-public class SageCaller extends TertiaryStage<SageOutput> {
+public abstract class SageCaller extends TertiaryStage<SageOutput> {
 
     public static final String SAGE_GENE_COVERAGE_TSV = "sage.gene.coverage.tsv";
     public static final String SAGE_BQR_PNG = "sage.bqr.png";
--- a/cluster/src/main/java/com/hartwig/pipeline/calling/structural/gripss/Gripss.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/calling/structural/gripss/Gripss.java
@@ -27,12 +27,13 @@ import com.hartwig.pipeline.report.ZippedVcfAndIndexComponent;
 import com.hartwig.pipeline.reruns.PersistedDataset;
 import com.hartwig.pipeline.reruns.PersistedLocations;
 import com.hartwig.pipeline.resource.ResourceFiles;
+import com.hartwig.pipeline.stages.Namespace;
 import com.hartwig.pipeline.stages.Stage;
 import com.hartwig.pipeline.storage.GoogleStorageLocation;
 import com.hartwig.pipeline.storage.RuntimeBucket;
 import com.hartwig.pipeline.tools.Versions;
 
-public class Gripss implements Stage<GripssOutput, SomaticRunMetadata> {
+public abstract class Gripss implements Stage<GripssOutput, SomaticRunMetadata> {
 
     private final InputDownload gridssVcf;
     private final InputDownload gridssVcfIndex;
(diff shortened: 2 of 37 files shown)


cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageCaller.java
```diff
@@ -28,7 +28,7 @@
 import com.hartwig.pipeline.storage.RuntimeBucket;
 import com.hartwig.pipeline.tertiary.TertiaryStage;
 
-public class SageCaller extends TertiaryStage<SageOutput> {
+public abstract class SageCaller extends TertiaryStage<SageOutput> {
```
pauldwolfe, line 31: I don't think this needs to be abstract

cluster/src/main/java/com/hartwig/pipeline/calling/structural/gripss/Gripss.java
```diff
@@ -27,12 +27,13 @@
...
 import com.hartwig.pipeline.reruns.PersistedLocations;
 import com.hartwig.pipeline.resource.ResourceFiles;
+import com.hartwig.pipeline.stages.Namespace;
 import com.hartwig.pipeline.stages.Stage;
 import com.hartwig.pipeline.storage.GoogleStorageLocation;
 import com.hartwig.pipeline.storage.RuntimeBucket;
 import com.hartwig.pipeline.tools.Versions;
 
-public class Gripss implements Stage<GripssOutput, SomaticRunMetadata> {
+public abstract class Gripss implements Stage<GripssOutput, SomaticRunMetadata> {
```
pauldwolfe, line 36: nor here

cluster/src/main/java/com/hartwig/pipeline/reruns/StartingPoint.java
```diff
@@ -50,9 +60,17 @@ enum StartingPoints {
                         Cobalt.NAMESPACE,
                         Amber.NAMESPACE,
                         SageConfiguration.SAGE_GERMLINE_NAMESPACE))),
-        GRIPSS_COMPLETE(concat(CALLING_COMPLETE.namespaces,
-                List.of(SOMATIC_NAMESPACE, GERMLINE_NAMESPACE))),
-        PURPLE_COMPLETE(concat(GRIPSS_COMPLETE.namespaces, List.of(Purple.NAMESPACE)));
+        GRIPSS_COMPLETE(concat(CALLING_COMPLETE.namespaces, List.of(SOMATIC_NAMESPACE, GERMLINE_NAMESPACE))),
+        PURPLE_COMPLETE(concat(GRIPSS_COMPLETE.namespaces, List.of(PaveSomatic.NAMESPACE, PaveGermline.NAMESPACE, Purple.NAMESPACE))),
```
pauldwolfe, line 64: I think it probably makes more sense to move PAVE up into CALLER_COMPLETE or its own starting point. Previously gene annotations were part of the somatic caller.

cluster/src/main/java/com/hartwig/pipeline/tertiary/cobalt/Cobalt.java
```diff
@@ -23,11 +23,15 @@
 import com.hartwig.pipeline.reruns.PersistedDataset;
 import com.hartwig.pipeline.reruns.PersistedLocations;
 import com.hartwig.pipeline.resource.ResourceFiles;
+import com.hartwig.pipeline.stages.Namespace;
 import com.hartwig.pipeline.storage.GoogleStorageLocation;
 import com.hartwig.pipeline.storage.RuntimeBucket;
 import com.hartwig.pipeline.tertiary.TertiaryStage;
 import com.hartwig.pipeline.tools.Versions;
 
+import jdk.jfr.Name;
```
pauldwolfe, line 32: unused

pauldwolfe (changes requested):

cluster/src/main/java/com/hartwig/pipeline/calling/sage/SageCaller.java
nedleitch: See comments below.

cluster/src/main/java/com/hartwig/pipeline/calling/structural/gripss/Gripss.java
nedleitch: I changed this to make it easier to distinguish between leaf nodes on the `Stage` class structure and those in the middle of the structure for the purposes of writing a useful test to check for classes missing the new `Namespace` annotation.

I had the same objection as you but then thought about whether instantiating this class directly would make sense and decided it wouldn't. This makes sense to me but maybe I'm missing something here. Let's discuss further so I can understand where you're coming from.

PR #407, opened 2022-05-19
DEV-2705 Remove Cram2Bam from protect-only start

nedleitch:


nedleitch:
Abandoning as specifying `-use_cram true` is more standard and requires no changes.

PR #411, opened 2022-06-07
DEV-2766 Separate Lilac BAM slicing from calling

nedleitch:


cluster/src/main/java/com/hartwig/pipeline/reruns/StartingPoint.java
```diff
@@ -66,21 +67,12 @@ enum StartingPoints {
                         LinxSomatic.NAMESPACE,
                         VirusAnalysis.NAMESPACE,
                         Chord.NAMESPACE,
+                        LilacBamSlicer.NAMESPACE,
```
pauldwolfe, line 70: I think we want to pull this up to an earlier a bit such that we can separate the bam slicing and lilac calling in reruns.

cluster/src/main/java/com/hartwig/pipeline/tertiary/lilac/Lilac.java
```diff
@@ -29,27 +27,32 @@
...
-    public Lilac(final AlignmentPair alignmentPair, final ResourceFiles resourceFiles, final PurpleOutput purpleOutput) {
-        super(alignmentPair);
+    public Lilac(final LilacBamSliceOutput slicedOutput, final ResourceFiles resourceFiles, final PurpleOutput purpleOutput) {
         this.resourceFiles = resourceFiles;
         PurpleOutputLocations purpleOutputLocations = purpleOutput.outputLocations();
         this.purpleGeneCopyNumber = initialiseOptionalLocation(purpleOutputLocations.geneCopyNumber());
         this.purpleSomaticVariants = initialiseOptionalLocation(purpleOutputLocations.somaticVariants());
+        this.slicedOutput = slicedOutput;
+        this.referenceBam = initialiseOptionalLocation(slicedOutput.reference());
+        this.tumorBam = initialiseOptionalLocation(slicedOutput.tumor());
```
pauldwolfe, line 55: do we still need both the sliced and non-sliced bams? I think part of the idea here was to make this stage not depend on anything but the slice bam

cluster/src/main/java/com/hartwig/pipeline/tertiary/lilac/LilacBamSlicer.java
```diff
@@ -0,0 +1,146 @@
...
+        return Stream.concat(buildCommands(getReferenceBamDownload().getLocalTargetPath(),
+                slicedBam(metadata.reference().sampleName())).stream(),
+                buildCommands(getTumorBamDownload().getLocalTargetPath(), slicedBam(metadata.tumor().sampleName())).stream())
+                .collect(toList());
+
+    }
+
+    @Override
+    public List<BashCommand> referenceOnlyCommands(final SomaticRunMetadata metadata) {
+        return buildCommands(getReferenceBamDownload().getLocalTargetPath(), metadata.sampleName());
```
pauldwolfe, line 60: why don't we slice for reference and tumor only?

cluster/src/main/java/com/hartwig/pipeline/tertiary/lilac/LilacBamSlicer.java
```diff
@@ -0,0 +1,146 @@
...
+        return VmDirectories.outputFile(sampleName + ".hla.bam");
+    }
+
+    private String bai(final String bam) {
+        return bam + ".bai";
+    }
+
+    private AddDatatype[] datatypes(final SomaticRunMetadata metadata) {
+        ArrayList<AddDatatype> datatypes = new ArrayList<>();
+        if (metadata.isSingleSample()) {
```
pauldwolfe, line 120: Maybe better to use the maybeReferenceSample and maybeTumorSample to decide whether to include the datatype.

pauldwolfe (changes requested):

cluster/src/main/java/com/hartwig/pipeline/tertiary/lilac/Lilac.java
nedleitch: Yeah this is just a naming thing, I'm caching the paths to the sliced BAMs as `referenceBam` and `tumorBam`, and keeping a reference to the full `slicedOutput` to access the indices (which I only need in one place). Could just make a member variable for the indices too but felt like it was polluting the namespace a bit.

pauldwolfe (approved):

PR #422, opened 2022-07-20
DEV-2805 Create VMs with provision mode SPOT

pauldwolfe:
To call this API needed to upgrade to the lastest version of the
compute API. Lots of changes but all good, builders and removal
of annoying "execute" methods.


nedleitch (approved):
Nice update!

PR #425, opened 2022-08-11
DEV-2829 Upgrade Protect to 2.3

nedleitch:


cluster/src/main/java/com/hartwig/pipeline/SomaticPipeline.java
```diff
@@ -210,6 +210,7 @@ public PipelineState run(final AlignmentPair pair) {
                                         linxSomaticOutput,
                                         virusInterpreterOutput,
                                         chordOutput,
+                                        lilacOutputFuture.get(),
```
pauldwolfe, line 213: this isn't the normal way to get the output of a stage future, should be done in the same place it is added to the state:
`ChordOutput chordOutput = pipelineResults.add(state.add(chordOutputFuture.get()));`

pauldwolfe (changes requested):

pauldwolfe (approved):

PR #427, opened 2022-08-12
DEV-2832 Remove termination action for local ssd/reduce amber footprint

pauldwolfe:
Removing the termination action entirely to allow for manual deletion of vms without
the pipeline restarting them.

Amber settings reduced as per Charles/Hongs recommendations and using a slightly
smaller thread count than available CPU to reduce saturation of VM.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/execution/vm/GoogleComputeEngine.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/execution/vm/GoogleComputeEngine.java
@@ -148,9 +148,6 @@ public class GoogleComputeEngine implements ComputeEngine {
 
                 if (arguments.usePreemptibleVms()) {
                     final Scheduling.Builder schedulingBuilder = Scheduling.newBuilder().setProvisioningModel("SPOT");
-                    if (arguments.useLocalSsds()) {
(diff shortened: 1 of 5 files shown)


cluster/src/main/java/com/hartwig/pipeline/execution/vm/GoogleComputeEngine.java
```diff
@@ -148,9 +148,6 @@ public PipelineStatus submit(final RuntimeBucket bucket, final VirtualMachineJob
 
                 if (arguments.usePreemptibleVms()) {
                     final Scheduling.Builder schedulingBuilder = Scheduling.newBuilder().setProvisioningModel("SPOT");
-                    if (arguments.useLocalSsds()) {
-                        schedulingBuilder.setInstanceTerminationAction("DELETE");
```
nedleitch, line 152: What happens in practice here? Will the Java library throw if this isn't set?

If I try to remove a VM with local SSDs via the Console it will display an error. I wonder if you added this to avoid a real problem you encountered?

pauldwolfe: Nothing is thrown actually. It will behave as it did before, the VM actually appears stopped but you can't start it up again.

PR #432, opened 2022-08-17
DEV-2836 Augment stage with addDatatypes method

nedleitch:
I've added this to the interface and also modified the implementors (with a few
exceptions) to override it and call it from both the output() and
persistedOutput() methdos.


cluster/src/main/java/com/hartwig/pipeline/stages/Stage.java
```diff
@@ -45,6 +46,10 @@ default S persistedOutput(final M metadata) {
         throw new UnsupportedOperationException(String.format("Stage [%s] does not support using persisted output.", namespace()));
     }
 
+    default List<AddDatatype> addDatatypes(final M metadata) {
```
pauldwolfe, line 49: I think the method should just be called `datatypes` as it doesn't actually add them to anything.

On that topic, I think we can take this a small step further such that stages don't need to explictly add the datatypes in their output methods. This could happen in the StageRunner, which could add the the datatypes to the output.

PR #434, opened 2022-08-18
DEV-2844 Serialize a dataset to json at end of run

pauldwolfe:
Adds all datatypes to a map to produce a compatible Dataset.java. The dataset is
serialized to json and uploaded to the output bucket.

### DIFF
--- a/cluster/src/main/java/com/hartwig/patient/Sample.java
+++ b/cluster/src/main/java/com/hartwig/patient/Sample.java
@@ -4,6 +4,8 @@ import java.util.List;
 import java.util.Optional;
 
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.hartwig.api.model.Run;
+import com.hartwig.api.model.RunSet;
 
 import org.immutables.value.Value;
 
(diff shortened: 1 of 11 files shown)


cluster/src/main/java/com/hartwig/patient/Sample.java
```diff
@@ -28,6 +30,9 @@ default String barcode() {
     }
 
     static ImmutableSample.Builder builder(final String name) {
+
+        Run run = new Run().id(1L).set(new RunSet().name("my set"));
```
nedleitch, line 34: Is this meant to stay this way?

cluster/src/main/java/com/hartwig/pipeline/transfer/staged/OutputDataset.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public void add(final AddDatatype datatype, final Blob blob) {
+        final Map<String, DatasetFile> sampleMap =
+                datasetMap.computeIfAbsent(datatype.dataType().toString().toLowerCase(), k -> new HashMap<>());
+        sampleMap.put(datatype.barcode(),
+                new DatasetFile().isDirectory(datatype.isDirectory()).path(blob.getName().replaceFirst(set, "").substring(1)));
+    }
+
+    public void serialize() {
```
nedleitch, line 33: I don't love the method name because it doesn't allude to the involvement of the bucket. Maybe mention the term "upload" or something?

nedleitch (changes requested):

PR #435, opened 2022-08-23
DEV-2853 Check for correct getopt before running

nedleitch:
This may be an anaemic check but it is better than what we had. On MacOS
the default getopt implementation is based on the BSD version and
doesn't support long options.

When invoked with "-V", the default version on modern Linux
installations (from the GNU foundation) returns a version, while the
MacOS version just returns "--". I've used this to create a simple
suitability check and we now print a warning with details for MacOS
users.

### DIFF
--- a/cluster/images/check_deps.sh
+++ b/cluster/images/check_deps.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+getopt -V | egrep -E '^getopty'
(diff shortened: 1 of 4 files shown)


pauldwolfe (approved):

stefvanlieshout (approved):
Script check_deps.sh works on my Mac (yields the getopt not as-expected msg).

PR #449, opened 2022-10-07
DEV-2922 Use samtools to support cram slicing

pauldwolfe:


cluster/src/main/java/com/hartwig/pipeline/tertiary/lilac/LilacBamSlicer.java
```diff
@@ -135,8 +138,15 @@ public boolean shouldRun(final Arguments arguments) {
...
-                new SambambaCommand("index", slicedBamFile));
+        return List.of(new SamtoolsCommand("view",
+                "-f",
+                FileTypes.isCram(inputBamFile) ? "cram" : "bam",
+                "-L",
+                resourceFiles.hlaRegionBed(),
+                "-b",
+                "-o",
+                slicedBamFile,
+                inputBamFile), new SambambaCommand("index", slicedBamFile));
```
nedleitch, line 149: I would very slightly prefer if we called the output `slicedBamOutputFile` or `slicedOutputBamFile` and the input file `inputFile` or `inputAlignedReads` or something as we don't know it's a BAM.

cluster/src/test/java/com/hartwig/pipeline/tertiary/lilac/LilacBamSlicerTest.java
```diff
@@ -31,9 +31,9 @@ protected Stage<LilacBamSliceOutput, SomaticRunMetadata> createVictim() {
     @Override
     protected List<String> expectedCommands() {
         return List.of(
-                "/opt/tools/sambamba/0.6.8/sambamba slice -L /opt/resources/lilac/37/hla.37.bed -o /data/output/reference.hla.bam /data/input/reference.bam",
+                "/opt/tools/samtools/1.14/samtools view -f bam -L /opt/resources/lilac/37/hla.37.bed -b -o /data/output/reference.hla.bam /data/input/reference.bam",
```
nedleitch, line 34: Maybe our current framework would make it too much trouble but it would be really great to test the conditional logic that decides which argument to pass for the input file type.

PR #455, opened 2022-11-21
DEV-2988 Port pipeline to new events api and module

pauldwolfe:
Package renaming and refactoring to use the new EventPublisher.

Switch to artifact repository for the events dependency.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/CommonArguments.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/CommonArguments.java
@@ -73,6 +73,8 @@ public interface CommonArguments {
 
     Optional<String> pubsubTopicWorkflow();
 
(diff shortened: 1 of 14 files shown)


cluster/src/main/java/com/hartwig/pipeline/CommonArguments.java
```diff
@@ -73,6 +73,8 @@ public interface CommonArguments {
 
     Optional<String> pubsubTopicWorkflow();
 
+    Optional<String> pubsubTopicEnvironment();
```
nedleitch, line 76: I know the precedent was there from before, but now that you've reworked the library I wonder if we need the "Topic" in the argument names?

nedleitch (approved):

PR #465, opened 2023-01-06
DEV-3036 Remove the dependency HMF API

pauldwolfe:
General idea is to move away from calling the api and always pass JSON in as pipeline input.

- Removes the dependency on the java client
- Removes all arguments and input related to SBP api
- Removes all the abstraction around the "metadata api"
- Makes the json input mandatory, and creates a PipelineInput object always at the start of the pipeline run.
- Datasets are no longer taken from the api, but also passed in as part of the "sample json"
- External ids (better name pending) are also optional in the sample json, and passed along such that they can be included in the final complete event

Still need to refactor a bit to make the event publishing optional.


nedleitch (approved):
Looks good, this is a really nice simplification!

As to the discussion on the ids, I don't like external either especially as the pipeline can be used by "external" parties. I like Stef's suggestions and will also add "job" and second the use of "hmf" somehow.

nedleitch (approved):

PR #467, opened 2023-01-12
DEV-2943 Remove sample info from WgsMetrics and BAM and CRAM

stefvanlieshout:
Remove all sample related info from the contents of output files (of diagnostic pipeline runs) that are re-used in research pipeline runs so that no such information "leaks" into the research context. This will allow to run diagnostic pipeline with different sample name.

- Removes sample info from ReadGroups
- Removes all PG lines from BAM header
- Removes all PG lines from CRAM header
- Removes all non-data content from WgsMetrics output

Note: the reheadering of BAM requires double the disk size in MergeMarkDup step because the output (from samtools reheader) is written to a new file (unfortunately sambamba does not allow for streaming output), so we might need to increase disk size somewhere?


cluster/src/main/java/com/hartwig/pipeline/execution/vm/unix/GrepFileCommand.java
```diff
@@ -0,0 +1,20 @@
+package com.hartwig.pipeline.execution.vm.unix;
+
+import static java.lang.String.format;
+
+import com.hartwig.pipeline.execution.vm.BashCommand;
+
+public class GrepFileCommand implements BashCommand {
```
pauldwolfe, line 7: GrepFileCommand could extend GrepCommand

cluster/src/main/java/com/hartwig/pipeline/execution/vm/unix/RedirectStdoutCommand.java
```diff
@@ -0,0 +1,20 @@
+package com.hartwig.pipeline.execution.vm.unix;
+
+import static java.lang.String.format;
+
+import com.hartwig.pipeline.execution.vm.BashCommand;
+
+public class RedirectStdoutCommand implements BashCommand {
```
pauldwolfe, line 7: Can we add a unit test for this guy?

pom.xml
```diff
@@ -70,10 +70,21 @@
                 <artifactId>events-pipeline</artifactId>
                 <version>${events.version}</version>
             </dependency>
+<!--            <dependency>-->
```
pauldwolfe, line 73: We can remove these as they didn't have an impact

pauldwolfe (changes requested):

cluster/src/main/java/com/hartwig/pipeline/execution/vm/unix/GrepFileCommand.java
stefvanlieshout: GrepCommand is no longer needed so removed.

pauldwolfe (approved):

cluster/src/main/java/com/hartwig/pipeline/execution/vm/unix/PipeCommands.java
```diff
@@ -6,7 +6,8 @@
 
 import com.hartwig.pipeline.execution.vm.BashCommand;
 
-public class PipeCommands implements BashCommand {
+public class
+PipeCommands implements BashCommand {
```
nedleitch, line 10: Looks like a mistake?

cluster/src/test/java/com/hartwig/pipeline/execution/vm/unix/DeleteFilesCommandTest.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.hartwig.pipeline.execution.vm.DeleteFilesCommand;
+
+import org.junit.Test;
+
+public class DeleteFilesCommandTest {
+    @Test
+    public void createsValidCommand() {
+        List<String> filesToDelete = Arrays.asList("file1", "file2");
+        assertThat(new DeleteFilesCommand(filesToDelete).asBash()).isEqualTo("rm file1 file2");
```
nedleitch, line 16: Inline?

cluster/src/test/java/com/hartwig/pipeline/execution/vm/unix/RedirectStdoutCommandTest.java
```diff
@@ -0,0 +1,20 @@
...
+import com.hartwig.pipeline.calling.command.SamtoolsCommand;
+
+import org.junit.Test;
+
+public class RedirectStdoutCommandTest {
+    @Test
+    public void createsValidCommand() {
+        String grepString = "view file.bam";
+        String outputFile = "/some/output/file";
+        assertThat(new RedirectStdoutCommand(new SamtoolsCommand(grepString), outputFile).asBash()).isEqualTo(format("(/opt/tools/samtools/1.14/samtools %s 1> %s)",
```
nedleitch, line 16: I do find this a bit confusing, and it introduces a dependency on `SamtoolsCommand`. Could we simplify it to just test the functionality the class under test provides?

nedleitch (changes requested):

cluster/src/test/java/com/hartwig/pipeline/execution/vm/unix/RedirectStdoutCommandTest.java
stefvanlieshout: Agreed and will replace with mock of BashCommand instead.

nedleitch (approved):

PR #469, opened 2023-02-01
Dev 3070

pauldwolfe:
Quick PR to leave some comments


cluster/src/main/java/com/hartwig/pipeline/PipelineMain.java
```diff
@@ -226,6 +236,16 @@ private PipelineCompleteEventPublisher createPublisher(final EventPublisher<Pipe
...
     }
 
+    private void apiUpdate(RunApi runApi, PipelineInput input, PipelineState state, Arguments arguments, String startOrFinish){
+        if (arguments.hmfApiUrl().isPresent()) {
+            Run run = runApi.get(input.operationalReferences().orElseThrow().runId());
+            if (startOrFinish.equals(START)){
+                HmfApiStatusUpdate.start(runApi, run);
+            } else if (startOrFinish.equals(FINISH)) {
+                HmfApiStatusUpdate.finish(runApi, run, state.status());
+            }
```
nedleitch, line 246: Looking at this critically for the first time I think I would dispense with the `String` constants and in fact this whole method and just call `HmfApiStatusUpdate.start` or `.finish` in the approprate place.

Also I think it would be better to have an instance of `HmfApiStatusUpdate` rather than calling static methods, I think we may have discussed this on the call we had, not sure. I'd pass the run to the constructor in that case.

cluster/src/main/java/com/hartwig/pipeline/metadata/HmfApiStatusUpdate.java
```diff
@@ -0,0 +1,50 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class HmfApiStatusUpdate {
+
+    private final static Logger LOGGER = LoggerFactory.getLogger(HmfApiStatusUpdate.class);
+    private static final String PIPELINE_SOURCE = "Pipeline";
+    private static final String HEALTH_CHECK = "HealthCheck";
+
+    public static void start(final RunApi runApi, final Run run) {
```
nedleitch, line 24: ...and probably the `RunApi` too could be passed to the constructor.

nedleitch (changes requested):

nedleitch (approved):

PR #472, opened 2023-02-17
Dev 3080

charlesshale:


cluster/src/main/java/com/hartwig/pipeline/SomaticPipeline.java
```diff
@@ -147,6 +147,7 @@ public PipelineState run(final AlignmentPair pair) {
                                     paveSomaticOutput,
                                     paveGermlineOutput,
                                     metadata.maybeTumor().map(t -> gripssSomaticProcessOutput).orElse(gripssGermlineProcessOutput),
```
mvanniekerkHartwig, line 149: Are you sure this line is correct? Because in purple this argument is exclusively referred to as `gripssSomaticOutput`, while according to this line it could also be `gripssGermlineProcessOutput` (if `metadata.maybeTumor` is empty).

mvanniekerkHartwig (changes requested):

cluster/src/main/java/com/hartwig/pipeline/SomaticPipeline.java
charlesshale: Yes have to think about that - that line has been there for a while and doesn't relate to my change, since I pass the Gripss germline SV VCF as the next argument. It relates to running the pipeline in germline/reference-only mode.

cluster/src/main/java/com/hartwig/pipeline/datatypes/DataType.java
```diff
@@ -5,24 +5,25 @@ public enum DataType {
...
+    GRIPSS_SOMATIC_STRUCTURAL_VARIANTS_UNFILTERED,
+    GRIPSS_SOMATIC_STRUCTURAL_VARIANTS,
+    GRIPSS_GERMLINE_STRUCTURAL_VARIANTS_UNFILTERED,
+    GRIPSS_GERMLINE_STRUCTURAL_VARIANTS,
+    PAVE_GERMLINE_VARIANTS,
+    PAVE_SOMATIC_VARIANTS,
+    PURPLE_GERMLINE_VARIANTS,
+    PURPLE_SOMATIC_VARIANTS,
+    PURPLE_SOMATIC_STRUCTURAL_VARIANTS,
+    PURPLE_GERMLINE_STRUCTURAL_VARIANTS,
```
nedleitch, line 26: These really do read better I think, but unfortunately this represents quite a big change as these enumeration constants are what we use to actually populate the datasets.

Put another way, while these are constants in the pipeline code and renaming them doesn't break anything here, their names are used to populate our API database. There won't be a non-zero downstream consequence here so we should be deliberate with changing these. Do they need to change?

cluster/src/test/java/com/hartwig/pipeline/tertiary/cuppa/CuppaTest.java
```diff
@@ -73,11 +73,17 @@ protected Stage<CuppaOutput, SomaticRunMetadata> createVictim() {
...
-                "/opt/tools/cuppa-chart/1.7.1_venv/bin/python /opt/tools/cuppa-chart/1.7.1/cuppa-chart.py " + "-sample tumor "
-                        + "-sample_data /data/output/" + TUMOR_CUP_DATA_CSV + " -output_dir /data/output",
-                "Rscript /opt/tools/cuppa/1.7.1/CupGenerateReport_pipeline.R tumor /data/output/");
+        return List.of("java -Xmx4G -jar /opt/tools/cuppa/1.7.2/cuppa.jar "
+                        + "-categories DNA "
+                        + "-ref_data_dir /opt/resources/cuppa/ "
+                        + "-sample_data tumor "
+                        + "-sample_data_dir /data/input/results "
+                        + "-output_dir /data/output "
+                        + "-create_pdf",
```
nedleitch, line 82: Annoyingly the code formatter settings will bring these all back to where they were to start with...

cluster/src/main/java/com/hartwig/pipeline/metrics/BamMetrics.java
```diff
@@ -81,24 +81,16 @@ public List<BashCommand> tumorReferenceCommands(final SingleSampleRunMetadata me
...
-                    intermediateFile));
-        }
-        bashCommands.add(new RedirectStdoutCommand(new GrepFileCommand(intermediateFile, "-A1 GENOME"), outputFile));
+        bashCommands.add(new BamMetricsCommand(
+                metadata.sampleName(),
+                bamDownload.getLocalTargetPath(),
+                resourceFiles,
+                VmDirectories.OUTPUT,
+                Bash.allCpus(),
+                arguments.useTargetRegions() ? resourceFiles.targetRegionsBed() : null));
```
nedleitch, line 92: `Optional.ofNullable()`...

nedleitch (changes requested):

cluster/src/main/java/com/hartwig/pipeline/tertiary/cuppa/Cuppa.java
```diff
@@ -87,31 +86,25 @@ public List<BashCommand> tumorOnlyCommands(final SomaticRunMetadata metadata) {
...
+        String cuppaOutputFile = VmDirectories.outputFile(format("%s.cup.data.csv", metadata.tumor().sampleName()));
+
+        List<String> chartArguments = Lists.newArrayList(
+                format("-sample %s", metadata.tumor().sampleName()),
+                format("-sample_data %s", cuppaOutputFile),
+                format("-output_dir %s", VmDirectories.OUTPUT));
+
+        return List.of(
+                new JavaJarCommand("cuppa", Versions.CUPPA, "cuppa.jar", "4G", cuppaArguments),
+                new Python3Command("cuppa-chart", Versions.CUPPA, "cuppa-chart.py", chartArguments));
```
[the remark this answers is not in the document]
nedleitch: I'm guessing not, this is from the commit that said Cuppa is now making its chart internally, correct?

cluster/src/main/java/com/hartwig/pipeline/SomaticPipeline.java
charlesshale: I've removed that condition since it should not have been there.

cluster/src/main/java/com/hartwig/pipeline/datatypes/DataType.java
charlesshale: They were already inconsistent as you can see and should remain changed. Sorry for the additional downstream work.

cluster/src/test/java/com/hartwig/pipeline/tertiary/cuppa/CuppaTest.java
charlesshale: Again it is clearer to have them on separate lines, as is done in other tests.

cluster/src/main/java/com/hartwig/pipeline/metrics/BamMetrics.java
```diff
@@ -56,9 +58,7 @@ public boolean shouldRun(final Arguments arguments) {
     }
 
     @Override
-    public List<BashCommand> inputs() {
-        return Collections.singletonList(bamDownload);
-    }
+    public List<BashCommand> inputs() { return ImmutableList.of(bamDownload, bamBaiDownload); }
```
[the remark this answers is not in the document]
charlesshale: Is there an issue with having a method like this on a single line?

cluster/src/main/java/com/hartwig/pipeline/metrics/BamMetricsCommand.java
```diff
@@ -0,0 +1,50 @@
...
+import com.hartwig.pipeline.resource.ResourceFiles;
+import com.hartwig.pipeline.tools.Versions;
+
+import org.jetbrains.annotations.Nullable;
+
+class BamMetricsCommand extends JavaClassCommand {
+
+    BamMetricsCommand(
+            final String sampleId, final String inputBam, final ResourceFiles resourceFiles, final String outputDir, final String threads,
+            @Nullable final String targetRegionsBed) {
```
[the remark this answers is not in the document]
charlesshale: Ok noted. 
Threads elsewhere is a string since Bash.cpus() returns a string for some reasons.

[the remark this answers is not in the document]
nedleitch: Yeah fair enough, I care less about the `Integer` and more about the `Optional`.

PR #474, opened 2023-02-21
Dev-3093: Update orange to v2.3, remove Rose and Protect from the molecular pipeline

mvanniekerkHartwig:
Builds off of DEV-3080


cluster/src/main/java/com/hartwig/pipeline/execution/vm/java/JavaJarFileExistsCommand.java
```diff
@@ -0,0 +1,26 @@
+package com.hartwig.pipeline.execution.vm.java;
+
+import com.hartwig.pipeline.execution.vm.BashCommand;
+
+public class JavaJarFileExistsCommand implements BashCommand {
```
nedleitch, line 5: Find the name a bit misleading here, but it's hard to think of a new one that isn't a mouthful.

Ideally I would prefer this not to be a `JavaJar...` command, but just a command that would emit some amount of code if the file existed. That could be called from within any other command as required. eg: 

```
public String asBash() {
  return String.format("[[ -e %s ]] && echo %s %s", targetPath, argument, targetPath");
}
```

Note that it probably doesn't event need to conform to the interface at all, though it is a convenient one in this case. Have a look at the `Bash.allCpus()` method and where it is used for maybe a clearer picture of what I am saying.

nedleitch (approved):

PR #479, opened 2023-03-28
DEV-3147 Stop/delete instances on startup

nedleitch:
May not be necessary with the changes from DEV-3146 but there is not a good reason I can think of not to cover as many bases as possible.


mvanniekerkHartwig (approved):

stefvanlieshout (approved):
Why do you want to stop a VM instead of delete in case of local SSD usage? As soon as the state is SUCCESS you should always be able to delete or not?

PR #482, opened 2023-04-05
DEV-3163 Disable ROSE and PROTECT for research

nedleitch:
These are out of date and for now it's easiest just to disable them.

I believe the smoke test will continue to function with no further changes and the unit tests also pass. We will likely be removing them soon anyway.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/tertiary/protect/Protect.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/tertiary/protect/Protect.java
@@ -188,6 +188,6 @@ public class Protect implements Stage<ProtectOutput, SomaticRunMetadata> {
 
     @Override
     public boolean shouldRun(final Arguments arguments) {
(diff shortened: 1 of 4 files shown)


mvanniekerkHartwig (commented):
`arguments.context() != Context.RESEARCH` would be a bit cleaner?

mvanniekerkHartwig (approved):

pauldwolfe (changes requested):
Could we update the tests to show that context is involved? (realize these tests aren't exhaustive, but we can at least include in the builder as docs)

```java
   @Override
    public void disabledAppropriately() {
        assertThat(victim.shouldRun(testDefaultsBuilder().runTertiary(false).shallow(false).build())).isFalse();
    }

    @Override
    public void enabledAppropriately() {
        assertThat(victim.shouldRun(testDefaultsBuilder().runTertiary(true).shallow(false).build())).isTrue();
    }
```

PR #486, opened 2023-04-21
DEV-3178 Further simplify VM management

nedleitch:
I have removed the messy loop that previously composed much of the executeSynchronously() method, replacing it with the mechanism recommended in Google's documentation for waiting on a running operation.

Logging of any errors in the InstanceLifecycleManager now occurs at the ERROR level, though most exceptions will be thrown back up to the StageRunner where the stage itself will be retried. This is somewhat of a middle ground; it can only result in more production alerts but should help us get a handle on which operations actually routinely fail, while also properly handling any errors that do come up.

There is more work to do here but I am reluctant to do too much as this is such a core piece of our infrastructure. I will revisit at some point in the future, when this release has been out in the wild for a while.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/execution/vm/GoogleComputeEngine.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/execution/vm/GoogleComputeEngine.java
@@ -63,18 +63,15 @@ public class GoogleComputeEngine implements ComputeEngine {
     private final Logger LOGGER = LoggerFactory.getLogger(GoogleComputeEngine.class);
 
     private final CommonArguments arguments;
-    private final ZonesClient zonesClient;
     private final ImagesClient images;
     private final Consumer<List<Zone>> zoneRandomizer;
     private final InstanceLifecycleManager lifecycleManager;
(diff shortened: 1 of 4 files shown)


pauldwolfe (changes requested):

cluster/src/main/java/com/hartwig/pipeline/execution/vm/GoogleComputeEngine.java
```diff
@@ -123,7 +121,7 @@ public PipelineStatus submit(final RuntimeBucket bucket, final VirtualMachineJob
             if (currentState == BucketCompletionWatcher.State.SUCCESS) {
                 LOGGER.info("Compute engine job [{}] already exists, and succeeded. Skipping job.", vmName);
                 lifecycleManager.findExistingInstance(vmName).ifPresent(instance -> {
-                    String zone = instance.getZone();
+                    String zone = instance.getZone().replaceAll(".*/", "");
```
pauldwolfe, line 124: not clear why this is necessary

pauldwolfe (approved):

PR #487, opened 2023-05-02
DEV-3227 Increase Purple memory

nedleitch:
There is currently a run in the research environment that is failing on a Java OutOfMemoryError. 31GB worked in the diagnostic run which is using an older Purple version.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/execution/vm/VirtualMachineJobDefinition.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/execution/vm/VirtualMachineJobDefinition.java
@@ -122,7 +122,7 @@ public interface VirtualMachineJobDefinition extends JobDefinition<VirtualMachin
                 .name("purple")
                 .startupCommand(bash)
                 .namespacedResults(resultsDirectory)
-                .performanceProfile(custom(4, 24))
+                .performanceProfile(custom(4, 32))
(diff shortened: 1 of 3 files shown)


cluster/src/main/java/com/hartwig/pipeline/execution/vm/VirtualMachineJobDefinition.java
```diff
@@ -122,7 +122,7 @@ static VirtualMachineJobDefinition purple(final BashStartupScript bash, final Re
                 .name("purple")
                 .startupCommand(bash)
                 .namespacedResults(resultsDirectory)
-                .performanceProfile(custom(4, 24))
+                .performanceProfile(custom(4, 32))
```
pauldwolfe, line 125: Are you sure this is enough if the heap does get to 31?

nedleitch: Summing up our offline conversation, no I'm not. But this change is currently only for the research environment and I would rather see what happens going to maximum memory for 4 cores than assume we're going to have a problem and up the count to 6 cores across the board.

In my testing with the research run I went from 18GB to 21GB to 24GB for the Purple Java process heap and none of these resolved the OOMError for the afflicted sample but it was fine with 31GB. There's a big gap between 24GB and 31GB so with any luck it doesn't get that close to 31GB and the OS never gets too close to running out at 32GB of system memory.

Either way should be an interesting test.

PR #491, opened 2023-05-08
DEV-3237 Upgrade orange to 2.4

pauldwolfe:
Version bump and add the disclaimer when RUO.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/SomaticPipeline.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/SomaticPipeline.java
@@ -8,6 +8,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 
+import com.hartwig.events.pipeline.Pipeline;
 import com.hartwig.pipeline.alignment.AlignmentPair;
 import com.hartwig.pipeline.calling.sage.SageGermlineCaller;
 import com.hartwig.pipeline.calling.sage.SageOutput;
@@ -219,7 +220,7 @@ public class SomaticPipeline {
                                         linxGermlineOutput,
                                         linxSomaticOutput,
                                         cuppaOutput,
-                                        virusInterpreterOutput, peachOutput, sigsOutput, resourceFiles)));
+                                        virusInterpreterOutput, peachOutput, sigsOutput, resourceFiles, arguments.context())));
                         Future<RoseOutput> roseOutputFuture = executorService.submit(() -> stageRunner.run(metadata,
                                 new Rose(resourceFiles,
                                         purpleOutput,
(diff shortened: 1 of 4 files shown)


cluster/src/test/java/com/hartwig/pipeline/tertiary/orange/OrangeTest.java
```diff
@@ -66,38 +73,63 @@ protected List<String> expectedInputs() {
...
+                TestInputs.linxGermlineOutput(),
+                TestInputs.linxSomaticOutput(),
+                TestInputs.cuppaOutput(),
+                TestInputs.virusInterpreterOutput(),
+                TestInputs.peachOutput(),
+                TestInputs.sigsOutput(),
+                TestInputs.REF_GENOME_37_RESOURCE_FILES,
+                Pipeline.Context.RESEARCH);
+        List<BashCommand> commands = victim.tumorReferenceCommands(TestInputs.defaultSomaticRunMetadata());
+        assertThat(commands.get(2).asBash()).contains("-add_disclaimer");
```
nedleitch, line 132: I don't love the use of the magic number, what about making it clear it's the last element using `commands.get(commands.size() - 1)`?

nedleitch (changes requested):

cluster/src/main/java/com/hartwig/pipeline/SomaticPipeline.java
```diff
@@ -219,7 +220,7 @@ public PipelineState run(final AlignmentPair pair) {
                                         linxGermlineOutput,
                                         linxSomaticOutput,
                                         cuppaOutput,
-                                        virusInterpreterOutput, peachOutput, sigsOutput, resourceFiles)));
+                                        virusInterpreterOutput, peachOutput, sigsOutput, resourceFiles, arguments.context())));
```
kduyvesteyn, line 223: Formatting looks pretty arbitrary!

cluster/src/main/java/com/hartwig/pipeline/tools/Versions.java
```diff
@@ -21,7 +21,7 @@ public interface Versions {
     String HEALTH_CHECKER = "3.4";
     String LILAC = "1.4.2";
     String LINX = "1.23.2";
-    String ORANGE = "2.3";
+    String ORANGE = "2.4";
```
kduyvesteyn, line 24: You should use 2.4.0

Related: I don't understand where the "2.4" jar comes from on data-vm, there is no "2.4" release.

kduyvesteyn (commented):
Apart from the comments in this review, there are some further requirements for being able to use ORANGE produced by pipeline in ACTIN (https://hartwigmedical.atlassian.net/browse/ACTIN-99)

cluster/src/test/java/com/hartwig/pipeline/tertiary/orange/OrangeTest.java
pauldwolfe: One way or another it's a magic number. I am actually looking for the third and not the second to last. I could also pluck out the orange command by filtering on the orange jar or something but then am i really adding value? As a compromise I'll assign to a descriptive variable such that its clear for others what we're doing here

PR #494, opened 2023-05-09
DEV-3242 Fix public image creation

nedleitch:
We discovered yesterday that we could no longer properly install the packages for Debian9 as they have moved them to an archive server and it seems have also changed some of the versions. I tried to get everything working using version pins and quickly found myself in dependency hell, and it just wasn't going to work.

Instead I created a VM from the last-successful image and deleted all the customisation (essentially just the files from `/opt`) and cut a new image from that in the GCP console. I applied the family `hmf-debian-9` to that and updated the `create_public_image.sh` script to reference it instead.

I've also commented out everything in the `*.cmds` files, I think this is fine (ie as opposed to deleting them) as we are likely to be dredging all the commands out again in a month or two anyway.


cluster/images/base.cmds
```diff
@@ -1,20 +1,20 @@
...
-sudo apt -y update
-sudo apt -y install perl-modules make time python3-pip python3-venv r-base vcftools openjdk-8-jdk bc
-sudo apt -y -t stretch-backports install openjdk-11-jdk
-sudo apt -y install libcurl4-gnutls-dev libxml2-dev libssl-dev less libgd-dev cpanminus parallel pigz git libmagick++-dev
-sudo apt -y install mdadm --no-install-recommends
-sudo pip3 install --no-cache-dir -U crcmod
-sudo mkdir /data
-sudo mkdir /opt/tools
-sudo mkdir /opt/resources
+#echo 'tmpfs /root/.gsutil tmpfs size=128m 0 0' | sudo tee -a /etc/fstab
```
pauldwolfe, line 2: I think the commenting will end up staying here and confusing someone down the line. Maybe write something as to why it was commented? Or put in a README?

nedleitch: Agreed, as I was pushing it this way I was thinking exactly these thoughts. I'll comment probably just to leave a breadcrumb back to the repo in `master` and then maybe backport to the other active branches.

PR #496, opened 2023-05-09
ACTIN-99 Add the second orange invocation

pauldwolfe:
Adds a boolean to toggle the germline to somatic conversion, and the namespace.

Run orange twice, once with the toggle on and once with it off.


cluster/src/main/java/com/hartwig/pipeline/tertiary/orange/Orange.java
```diff
@@ -51,9 +51,10 @@
 import com.hartwig.pipeline.tertiary.virus.VirusInterpreterOutput;
 import com.hartwig.pipeline.tools.Versions;
 
-@Namespace(Orange.NAMESPACE)
+@Namespace(Orange.NAMESPACE_NO_GERMLINE)
```
pauldwolfe, line 54: This is an intellij refactor gone wrong but @nedleitch how this annotation actually used? Is it possible to have 2 namespaces for one stage or do i need to subclass?

nedleitch: That's certainly not the intended usage but clearly the annotation does not enforce this. The idea was that these annotations would replace the constants that most of them are referencing, ultimately strengthening the contract on the stages and paving the way for a declarative stage runner. We didn't really push too hard so now its value is limited.

It _is_ involved with automated testing to ensure that all stages are mentioned in at least one starting point. That was added for the Protect reruns to ensure that when a new stage is added it won't get run by that starting point, which we'd hit if you were to subclass. Again maybe the value is limited.

nedleitch (approved):

cluster/src/main/java/com/hartwig/pipeline/input/MetadataProvider.java
```diff
@@ -33,6 +33,7 @@ public SomaticRunMetadata get() {
                         .barcode(barcodeOrSampleName(t))
                         .sampleName(t.name())
                         .primaryTumorDoids(t.primaryTumorDoids())
+                        .samplingDate(t.samplingDate())
```
kduyvesteyn, line 36: What's the process for populating the SampleInput? 

I assume `samplingDate` is an optional field also in the SampleInput?

cluster/src/main/java/com/hartwig/pipeline/tertiary/orange/Orange.java
```diff
@@ -255,7 +261,7 @@ private List<BashCommand> buildCommands(final SomaticRunMetadata metadata) {
     @Override
     public VirtualMachineJobDefinition vmDefinition(final BashStartupScript bash, final ResultsDirectory resultsDirectory) {
         return VirtualMachineJobDefinition.builder()
-                .name(NAMESPACE)
+                .name(NAMESPACE_NO_GERMLINE)
```
kduyvesteyn, line 264: Shouldn't this be `namespace()` rather than the hardcoded NO_GERMLINE?

cluster/src/test/java/com/hartwig/pipeline/tertiary/orange/OrangeTest.java
```diff
@@ -99,7 +100,7 @@ protected List<String> expectedCommands() {
                         + "-cohort_percentiles_tsv /opt/resources/orange/cohort_percentiles.tsv "
                         + "-driver_gene_panel_tsv /opt/resources/gene_panel/37/DriverGenePanel.37.tsv "
                         + "-known_fusion_file /opt/resources/fusions/37/known_fusion_data.37.csv "
-                        + "-ensembl_data_directory /opt/resources/ensembl_data_cache/37/ " + "-convert_germline_to_somatic";
+                        + "-ensembl_data_directory /opt/resources/ensembl_data_cache/37/ -experiment_date 2023-05-19";
```
kduyvesteyn, line 103: This experiment date will give a WARN in ORANGE, since only supported format is yyMMdd.

cluster/src/main/java/com/hartwig/pipeline/tertiary/orange/Orange.java
```diff
@@ -238,11 +241,14 @@ private List<BashCommand> buildCommands(final SomaticRunMetadata metadata) {
...
-                        resourceFiles.ensemblDataCache(),
-                        "-convert_germline_to_somatic");
+                        resourceFiles.ensemblDataCache());
         if (context.equals(Pipeline.Context.RESEARCH)) {
             argumentListBuilder.add("-add_disclaimer");
         }
+        if (!includeGermline) {
+            argumentListBuilder.add("-convert_germline_to_somatic");
+        }
+        metadata.tumor().samplingDate().ifPresent(sd -> argumentListBuilder.add("-experiment_date", sd.toString()));
```
kduyvesteyn, line 251: I suggest to reformat to `yyMMdd` since any other format will give a WARN in ORANGE

kduyvesteyn (commented):
The only real functional issue I could find relates to the formatting of the `samplingDate` to the format expected by ORANGE. In addition, some questions in various places.

cluster/src/main/java/com/hartwig/pipeline/input/MetadataProvider.java
pauldwolfe: It is the responsibility of the launcher of the pipeline to pass it in as config (same with the doids). In the end it comes from LAMA, same field as is currently extracted in the actin script.

cluster/src/main/java/com/hartwig/pipeline/tertiary/orange/Orange.java
pauldwolfe: it should!

cluster/src/test/java/com/hartwig/pipeline/tertiary/orange/OrangeTest.java
pauldwolfe: Ok, it would be best if we stuck with ISO convention date formats but probably not worth an ORANGE release. I'll update.

kduyvesteyn: It would be nice if you could update ORANGE nonetheless I think, just don't make a release yet. 

In addition to supporting ISO, I think it would be better to throw an exception if someone explicitly passed a date in a format that is not supported (reverting to `now` feels arbitrary)

```
    @NotNull
    private static LocalDate interpretExperimentDateParam(@NotNull String experimentDateString) {
        String format = "yyMMdd";

        LocalDate experimentDate;
        try {
            experimentDate = LocalDate.parse(experimentDateString, DateTimeFormatter.ofPattern(format, Locale.ENGLISH));
            LOGGER.debug("Configured experiment date to {}", experimentDate);
        } catch (DateTimeParseException exception) {
            experimentDate = LocalDate.now();
            LOGGER.warn("Could not parse configured experiment date '{}'. Expected format is '{}'", experimentDateString, format);
        }
        return experimentDate;
    }
```

PR #511, opened 2023-08-18
DEV-3541 Don't add datatypes to germline Orange

nedleitch:
These files are used only by ACTIN currently and that project doesn't need the datatypes at the moment. Without this change we end up with a list of items under the key in the dataset which breaks existing clients.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/tertiary/orange/Orange.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/tertiary/orange/Orange.java
@@ -33,7 +33,6 @@ import com.hartwig.pipeline.tertiary.virus.VirusInterpreterOutput;
 import com.hartwig.pipeline.tools.VersionUtils;
 
 import java.time.format.DateTimeFormatter;
-import java.util.Collections;
 import java.util.List;
 
 import static com.hartwig.pipeline.execution.vm.InputDownload.initialiseOptionalLocation;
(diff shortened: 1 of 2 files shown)


cluster/src/main/java/com/hartwig/pipeline/tertiary/orange/Orange.java
```diff
@@ -270,11 +257,15 @@ public OrangeOutput persistedOutput(final SomaticRunMetadata metadata) {
 
     @Override
     public List<AddDatatype> addDatatypes(final SomaticRunMetadata metadata) {
-        final String orangePdf = metadata.tumor().sampleName() + ORANGE_OUTPUT_PDF;
-        final String orangeJson = metadata.tumor().sampleName() + ORANGE_OUTPUT_JSON;
-        return List.of(new AddDatatype(DataType.ORANGE_OUTPUT_JSON,
-                        metadata.barcode(),
-                        new ArchivePath(Folder.root(), namespace(), orangeJson)),
-                new AddDatatype(DataType.ORANGE_OUTPUT_PDF, metadata.barcode(), new ArchivePath(Folder.root(), namespace(), orangePdf)));
+        if (includeGermline) {
```
pauldwolfe, line 260: This is the wrong way around. We want to attach the datatypes to the complete orange output with germline.

pauldwolfe (changes requested):

pauldwolfe (approved):

PR #512, opened 2023-08-22
DEV-3508 Refactor starting points for 5.33 rerun

nedleitch:
The desired steps for the 5.33 rerun are the same as the set for the 5.32 were so I decided to just make a new permanent starting point for this situation, and have called it `PAVE_GRIPSS_ONWARD`.

I noticed there was quite some duplication between some of the existing sets so eliminated that. I also decided to just leave in the non-germline Orange invocation that was added for ACTIN as Orange runs very quickly and it seemed likely the ACTIN pipeline might be rerun at some point in the future.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/reruns/StartingPoint.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/reruns/StartingPoint.java
@@ -32,53 +32,6 @@ public class StartingPoint {
 
     private final StartingPoints startingPoint;
 
-    enum StartingPoints {
-        BEGINNING(Collections.emptyList()),
-        ALIGNMENT_COMPLETE(List.of(Aligner.NAMESPACE, BamMetrics.NAMESPACE, Flagstat.NAMESPACE, SnpGenotype.NAMESPACE)),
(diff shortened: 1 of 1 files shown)


cluster/src/main/java/com/hartwig/pipeline/reruns/StartingPoint.java
```diff
@@ -32,53 +32,6 @@ public class StartingPoint {
...
-
-        RERUN_532(concat(SKIP_GRIDSS.namespaces,
-                List.of(Cobalt.NAMESPACE,
-                        Amber.NAMESPACE,
-                        SageConfiguration.SAGE_GERMLINE_NAMESPACE,
-                        SageConfiguration.SAGE_SOMATIC_NAMESPACE,
-                        VirusBreakend.NAMESPACE))),
-
-        RESEARCH_AFTER_531_DIAGNOSTIC(concat(ALIGNMENT_COMPLETE.namespaces, List.of(GermlineCaller.NAMESPACE))),
-        BWA_COMPLETE(List.of(Aligner.NAMESPACE));
```
nedleitch, line 69: `BWA_COMPLETE` was temporary, added only for verification of the changes in 5.33 to use new tooling for the BAM metrics and flagstats. Should not be used again.

PR #514, opened 2023-08-31
ACTIN-288: Bump ORANGE to v2.7.0

kduyvesteyn:
It feels too easy to me so please check if I haven't omitted a step somewhere!


pauldwolfe (commented):
Weren't there some argument changes we needed to make along with the version bump?

We do need to make the actual image, which is done by running the `cluster/images/create_public_image.sh` once the jar has been uploaded to the bucket.

pauldwolfe:
I've created the image and rerun the smoke test, it all passes so I suppose we didn't need any different arguments. Will merge this and upgrade

kduyvesteyn:
Thanks. Sorry, should have mentioned there are no argument changes.

PR #519, opened 2023-09-07
Aus 165 - remove HLA BAM slice

charlesshale:


charlesshale:
Smoke tests pass

pauldwolfe (commented):
Changes look good, but neither PR commits nor ticket mention why we're removing this? As there is no update to Lilac assume that we're going to lose whatever performance improvement on rerun we got from staging the bam slicing?

Also com.hartwig.pipeline.tertiary.lilac.LilacTest is failing.

charlesshale:
Fixed unit test and updated ticket with rationale

PR #522, opened 2023-09-20
DEV-3595 Remove JSON credentials handling

nedleitch:
Platinum and the production environment have already been changed to drop the use of the service account keys. The only question is how the smoke test will go without the JSON files, could be some trial and error getting it working inside the build environment.

Really sorry about the formatting, I reimported the code formatter settings beforehand and my Intellij is set to run the formatter on file save, so either there's something funny about my Intellij itself or I've done something silly...


KasperWolsink99 (approved):

cluster/src/test/java/com/hartwig/pipeline/smoke/SmokeTest.java
```diff
@@ -121,14 +152,10 @@ public void runFullPipelineAndCheckFinalStatus(final String inputMode, final Pip
                 .outputBucket("smoketest-pipeline-output-pilot-1")
                 .useTargetRegions(false)
                 .refGenomeVersion(refGenomeVersion);
-
-        if (whoami.equals("root")) {
-            String privateKeyPath = workingDir() + "/google-key.json";
-            builder.privateKeyPath(privateKeyPath).uploadPrivateKeyPath(privateKeyPath);
-        }
+        //                .serviceAccountEmail("pipeline5-build@hmf-build.iam.gserviceaccount.com");
```
mvanniekerkHartwig, line 155: Should you get rid of this line?

mvanniekerkHartwig (approved):
Bit hard to review but looks ok...

cluster/src/test/java/com/hartwig/pipeline/smoke/SmokeTest.java
nedleitch: Thank-you. Removed.

PR #523, opened 2023-09-29
DEV-3490 update release method

KasperWolsink99:
- Update cloud build to use `docker-tag` and `$TAG_NAME`.
- Add mvn `extension.xml` file to ensure ide support for downloading sources and documentation.
- Change deployment location to artifact registry.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -46,6 +46,6 @@ steps:
     dir: cluster
     args: ['eu.gcr.io/hmf-build/pipeline5', '$TAG_NAME']
 images:
-  - eu.gcr.io/$PROJECT_ID/pipeline5
+  - eu.gcr.io/hmf-build/pipeline5
 logsBucket: 'gs://hmf-build-logs'
 timeout: 4800s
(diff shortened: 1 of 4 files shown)


pauldwolfe (approved):

cloudbuild.yaml
```diff
@@ -8,7 +8,7 @@ steps:
   - name: 'maven:3-jdk-11'
     id: 'Set version for Maven'
     entrypoint: mvn
-    args: [ 'versions:set', '-DnewVersion=${_MAJOR_MINOR_VERSION}.$SHORT_SHA' ]
+    args: [ 'versions:set', '-DnewVersion=${TAG_NAME}' ]
```
nedleitch, line 11: My only concern with this is whether it makes it easier to build a pipeline that doesn't match the branch it is on. Pipeline5 is the only project (to my knowledge) that we build and release from multiple branches concurrently.

Not a question of the correctness of what you've done here technically, more a strategic question.

nedleitch (approved):

cloudbuild.yaml
KasperWolsink99: Will discuss this with the Amsterdam people

nedleitch: Yeah agreed. I think that's a good way to move forward without causing too much extra work.

PR #525, opened 2023-10-26
DEV-3663: Update orange to v3.0.0

kzuberihmf:
* Updates orange to v3.0.0
* Adds `-experiment_type WGS` parameter to orange, hardcoded for now
* Changes `-experiment_date` to `-sampling_date`
* Removes creation of linx_plot_dir, Orange should work without this
* Small updates to pilot image testing documentation

Supersedes DEV-3650 which was merged into this pr.


cluster/src/main/java/com/hartwig/pipeline/tertiary/orange/Orange.java
```diff
@@ -152,9 +152,13 @@ private List<BashCommand> buildCommands(final SomaticRunMetadata metadata) {
         final List<String> primaryTumorDoids = metadata.tumor().primaryTumorDoids();
         String primaryTumorDoidsString = "\"" + String.join(";", primaryTumorDoids) + "\"";
         String linxPlotDir = linxSomaticOutputDir.getLocalTargetPath() + "/plot";
+        // TODO can we determine experiment type from pipeline metadata? Can be WGS or PANEL, hardcode to WGS for now
```
pauldwolfe, line 155: Since this "buildCommands" is only being called by tumorReferenceCommands, you have essentially already determined that this is a WGS experiment type, as we do not run targeted with both tumor and reference. But indeed that's a bit implicit.

You could use the `Arguments.useTargetRegions()` to more explicitly discriminate.

Regardless of what we decide, lets remove the TODO, we generally don't have them in this code base. You can mention in the commit message instead to have some history/rationale.

pauldwolfe (changes requested):

pauldwolfe (approved):

kduyvesteyn (approved):
Got no further comments other than the fact that ORANGE should become v3.0.1

Also, it's weird that the pipeline version is present in-memory but we have to write it to a file for ORANGE to pick up. Better to provide it directly as an argument. Can leave for future improvement perhaps.

pauldwolfe (approved):

PR #528, opened 2023-11-01
DEV-3676 Bump Purple heap to 31G

nedleitch:
This was done in 5.32 but didn't get propagated to 5.33.

When this has been approved it will also be merged to master.


stefvanlieshout (approved):
You say "will also be merged into master" but assume that means "will also be merged into 5.33-hotfix"?

PR #531, opened 2023-11-08
DEV-3688 Unify VM core and memory declarations

nedleitch:
The `HmfTool` class and the VM definitions both define parameters for the VMs that our in-house tooling runs on. I noticed this while trying to increase the memory for Purple and when it wasn't working realised the situation.

Reference the HmfTool-hosted constants from within the VM definition calls whenever they exist as not doing this violates the principle of least surprise. Also I think the intent of the HmfTool class must have been to centralise the configuration as much as possible and this is a step in that direction.

I am aware this is an imperfect solution as the constants don't exist for those tools we don't maintain and a couple of the tools use different parameters depending on which "mode" they've been called in, Sage springs to mind.

I checked pretty thoroughly for references and I think I got them all but I suppose we'll find out in the SmokeTest if anything is off.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/execution/vm/VirtualMachineJobDefinition.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/execution/vm/VirtualMachineJobDefinition.java
@@ -1,7 +1,16 @@
 package com.hartwig.pipeline.execution.vm;
 
 import static com.hartwig.pipeline.execution.vm.VirtualMachinePerformanceProfile.custom;
+import static com.hartwig.pipeline.tools.HmfTool.AMBER;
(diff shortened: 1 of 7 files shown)


nedleitch:
Sorry for the formatting here, I don't understand why my Intellij keeps shuffling things around. I thought I had disabled all of the places where it would do that but clearly I've missed something. Open to suggestions.

stefvanlieshout (approved):
Yeah the formatting changes make it a bit hard to follow, but having all config per tool in `HmfTool.java` make it very easy to grasp, so nice one :) Not sure if you also want to merge into `533-hotfix` branch, then fyi the PR is now set to merge into `master`.

Also, Kasper made changes (in a separate branch) to extract compute logic into hmf-cloud-sdk so that might start to clash with some of the changes in here?

PR #533, opened 2023-11-09
DEV-3483 Demote retried operation logging to warn

nedleitch:
The changes that resulted in adding the retries have been battle-tested in production and I'm confident we've got a correct implementation so we can avoid alerting on it every time now.


stefvanlieshout (approved):
Thanks, this will save time every new sequencing rounds from now on!

PR #538, opened 2023-12-13
INFRA-119 refactor pipeline5 to use hmf-cloud-sdk execution.vm

KasperWolsink99:
- Remove the `execution.vm` package from this repository (alongside some direct dependencies) and move them to the `hmf-cloud-sdk` repository and use this instead.


cluster/src/main/java/com/hartwig/pipeline/Arguments.java
```diff
@@ -1,13 +1,11 @@
 package com.hartwig.pipeline;
 
-import java.util.Optional;
```
mvanniekerkHartwig, line 3: You're not using the `hmf_code_scheme` from the root of this repo. That is causing a lot of the lines changed. Can you go through all changes in the PR and make sure the correct formatting is applied?

mvanniekerkHartwig (changes requested):
Please fix the formatting to reduce LOC changed.

cluster/src/main/java/com/hartwig/pipeline/Arguments.java
KasperWolsink99: undid the formatting commits, it should mostly be gone now.

PR #541, opened 2023-12-27
DEV-3720 Imager changes to use Artifact Registry

nedleitch:
In the fullness of time this will replace the `common-tools` bucket entirely for HMF-maintained tools. For the moment it simply provides an alternative flow to allow fetching artifacts from our internal Artifact Registry. As tools are made available inside the registry they can be removed from the bucket and added to the list in `VersionUtils` to be fetched from the registry instead.


image/fetch_tool_from_registry.sh
stefvanlieshout, line 1: This will be a very useful script to have in general.

cluster/src/main/java/com/hartwig/pipeline/tools/VersionUtils.java
```diff
@@ -50,6 +51,10 @@ public static String majorMinorVersion(final String version) {
         return version;
     }
 
+    public static List<HmfTool> inArtifactRegistry() {
+        return List.of(HmfTool.ORANGE);
```
stefvanlieshout, line 55: Assuming this means that ony ORANGE is now retrieved from repo instead of bucket it contradicts your statement in readme that all HMF tools should now be pushed to repo, but maybe adding the rest comes later, in that case fine of course.

stefvanlieshout (approved):

PR #543, opened 2024-01-03
DEV-3730 Updates to imager for Debian 12

nedleitch:


nedleitch:
Replaced with INFRA-250 per comments.

PR #546, opened 2024-01-15
INFRA-189 Update Compar expectations

nedleitch:
I think this is all that is required, I have a green build on my machine with these updates.

Would appreciate somebody who can properly qualify the changes I had to make from a biological perspective sign off on these amendments.


stefvanlieshout (approved):
All changes make sense (double checked with Erik) and I also succesfully ran the smoke test locally so approved!

PR #558, opened 2024-03-13
INFRA-326 Allow usage of Cuppa private resource file

stefvanlieshout:
Makes it possible for pipeline5 to run in both a public mode (using a public compute image without the private CUPPA resource file) and private mode (using a private compute image with the private CUPPA resource file)

For this I added argiment `use_private_resources`. The only downside is that to actually use the private resource file (cuppa_cv_predictions.37.tsv.gz) the argument must be provided, but this makes the intention also very clear and explicit so worth the trade-off.

Previous code tried to detect the existance of the file but that did not work since that check was executed in the k8 job instead of on the CUPPA stage VM.


KasperWolsink99 (approved):
lgtm

mvanniekerkHartwig (approved):

PR #561, opened 2024-04-02
ACTIN-788 Make new pipeline5 release containing (at least) CUPPA v2.1.0

nedleitch:
Pull in recent changes from the 5.34 hotfix branch (except for most imager modifications as `master` is already on Debian 12 and they are not compatible).


README.md
```diff
@@ -132,21 +132,14 @@ rules](https://chris.beams.io/posts/git-commit/) and follow them.
 
 ### CI 
 
-After each push our CI build runs on Google's Cloud Build in our development project. In addition to the tests you might run
-locally the CI build also:
+There is a CI environment that will test the pipeline using cut-down COLO829 "mini" FASTQs against the most-recent disk image
+available for the version of the pipeline. This is usually initiated by pushing a `beta` tag to the PR branch being modified.
```
pauldwolfe, line 136: The smoke test runs on every push to the PR, a tag isn't required.

cluster/src/main/java/com/hartwig/pipeline/Arguments.java
```diff
@@ -211,6 +215,8 @@ static ImmutableArguments.Builder defaultsBuilder(final String profileString) {
 
     boolean anonymize();
 
+    boolean usePrivateResources();
```
pauldwolfe, line 218: This is a little awkward. Now users need to remember to have this flag set when they use the private image, and its not clear that this is very specific to cuppa. Could we think of a way to do it dynamically based on whether the private resource exists? Or have an empty one in the public image to remove the optionality?

image/README.md
```diff
@@ -1,11 +1,13 @@
 # What is this directory?
 
-Tools for creating new disk images for running PipelineV5 VMs on GCP. The image must include all of the tools that are required by
-the pipeline version being executed, and must be a functional standalone OS. The images we create are based on the Debian image
-provided by Google. There are multiple aspects to this image:
-
-* Our tools or their dependencies use Python, R and Perl. As of the update to Debian 12 the latter two of these are managed
-  via `anaconda`.
+Tools for creating new disk images for running `pipeline5` VMs on GCP. The `pipeline5` application consists of a Java application
```
pauldwolfe, line 3: These README edits don't relate to CUPPA 2.1.0. Can we move this to a separate ticket/PR?

image/create_private_image.sh
```diff
@@ -0,0 +1,176 @@
+#!/usr/bin/env bash
```
pauldwolfe, line 1: Also unrelated to CUPPA 2.1.x. Did we somehow lose history of this file?

pauldwolfe (changes requested):
Also, smoke test failing on a missing file passed to orange. Do we need ORANGE changes to properly handle the output? If so I guess we should also check expected_outputs of the smoke test?

image/create_private_image.sh
nedleitch: All of these came in from 5.34-hotfix. As mentioned in the commit messages I retained the changes made to the imaging scripts but the rest has come from there.

image/README.md
nedleitch: This was addressing the request for improved documentation in a separate email. I can split them out but maybe more pragmatic just to modify the ticket name.

nedleitch:
Abandoning this because all of the functionality has been implemented in more-targeted tickets.

PR #562, opened 2024-04-02
ACTIN-799 Improve Anaconda automation

nedleitch:
This was originally about adding the Perl Soundex module to the Anaconda-managed environment but it has turned into something a little more.

The part of the feature to upload the finished tarball to the bucket may actually be unnecessary with the improvements made to the performance by switching to `mamba`. I am waiting for some performance numbers to confirm this, removing that feature
would be preferable to trying to figure out how to manage the tarballs frankly.


nedleitch:
The imaging run just-finished took 38 minutes which is I think quite reasonable.

On the strength of that I'm going to set about removing the extra bits that upload the completed tarball to the bucket, this will reduce the time further.

PR #564, opened 2024-04-04
ACTIN-809 Reorder stage management statements

nedleitch:
Teal does not need to hold the whole pipeline execution hostage because its output is not used anywhere. On the run I was watching, Teal ran for exactly 1 hour which is significant.

I also moved Lilac down to just above the "final" block of quick- running tools at the end since it too was running synchronously.


cluster/src/main/java/com/hartwig/pipeline/SomaticPipeline.java
```diff
@@ -62,10 +54,13 @@
 import com.hartwig.pipeline.tertiary.virus.VirusBreakendOutput;
 import com.hartwig.pipeline.tertiary.virus.VirusInterpreter;
 import com.hartwig.pipeline.tertiary.virus.VirusInterpreterOutput;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.concurrent.*;
```
pauldwolfe, line 60: star import

pauldwolfe (approved):

PR #565, opened 2024-04-05
ACTIN-818 Bump MarkDups to 1.1.4

nedleitch:
Also revise imaging script behaviour to print out a message with details on where the script was generated to, rather than require the user to copy/paste commands to get it to run.


kduyvesteyn (approved):
No strong opinion on the changes to public image creation

PR #566, opened 2024-04-09
ACTIN-835 Fix minor recent Pv5 merge mistakes + use AR

nedleitch:
Probably not super-exciting, much of this is just restoring some stuff that disappeared from `master` sometime in February I think it was.

Once the changes were restored I decided to update the list of JARs that is coming from Artifact Registry too, when I first made these changes only Orange was in there but now we've got a decent-sized list. Think it would be worthwhile to continue this process and eventually mostly eliminate the `hmftools` bucket from the flow for HMF-maintained tooling.

### DIFF
--- a/image/create_public_image.sh
+++ b/image/create_public_image.sh
@@ -7,6 +7,8 @@ ZONE="${LOCATION}-a"
 PROJECT="hmf-pipeline-development"
 PV5_DIR="$(dirname "$0")/.."
 PV5_JAR="${PV5_DIR}/cluster/target/cluster-local-SNAPSHOT.jar"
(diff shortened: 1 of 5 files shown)


image/create_public_image.sh
```diff
@@ -39,10 +41,15 @@ done
 set -e
 echo "Rebuilding pipeline JAR to ensure correct version"
 mvn -f "${PV5_DIR}/pom.xml" clean package -DskipTests
-version="$(java -cp ${PV5_JAR} com.hartwig.pipeline.tools.VersionUtils)"
+version="$($VERSION_CMD)"
```
pauldwolfe, line 44: this is cool :)

pauldwolfe (approved):

PR #568, opened 2024-04-26
INFRA-75 Support RESEARCH2 context

mdenburger:
Supports the new RESEARCH2 context that runs a second research pipeline in parallel with the normal RESEARCH pipeline run. The RESEARCH2 run will use a different set name and different sample names and barcodes.


KasperWolsink99 (approved):
lgtm

stefvanlieshout (approved):
I think some of the (existing) logic could be simplified, for example the checks to determine if healthchecker should run or not. But out of scope for this PR so approving.

PR #569, opened 2024-05-09
INFRA-469: when a gcp request has failed, print the causes

hongwingl:
This is to fix a problem where gcp request failure just prints a `Bad request` with no reason given.

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/failsafe/DefaultBackoffPolicy.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/failsafe/DefaultBackoffPolicy.java
@@ -20,9 +20,16 @@ public class DefaultBackoffPolicy<R> extends RetryPolicy<R> {
         abortOn(InvalidArgumentException.class);
         onAbort(e -> LOGGER.error("Unable to submit operation", e.getFailure()));
         handle(Exception.class);
-        onFailedAttempt(rExecutionAttemptedEvent -> LOGGER.warn("[{}] failed: {}",
-                taskName,
-                rExecutionAttemptedEvent.getLastFailure().getMessage()));
+        onFailedAttempt(rExecutionAttemptedEvent ->
+        {
+            // we need to keep tracing the cause to print out the real failure reason
+            Throwable lastFailure = rExecutionAttemptedEvent.getLastFailure();
+            while(lastFailure != null)
+            {
+                LOGGER.warn("[{}] failed: {}", taskName, lastFailure.getMessage());
+                lastFailure = lastFailure.getCause();
+            }
+        });
     }
 
     public static <R> DefaultBackoffPolicy<R> of(final String taskName) {


cluster/src/main/java/com/hartwig/pipeline/failsafe/DefaultBackoffPolicy.java
```diff
@@ -20,9 +20,16 @@ public class DefaultBackoffPolicy<R> extends RetryPolicy<R> {
         abortOn(InvalidArgumentException.class);
         onAbort(e -> LOGGER.error("Unable to submit operation", e.getFailure()));
         handle(Exception.class);
-        onFailedAttempt(rExecutionAttemptedEvent -> LOGGER.warn("[{}] failed: {}",
-                taskName,
-                rExecutionAttemptedEvent.getLastFailure().getMessage()));
+        onFailedAttempt(rExecutionAttemptedEvent ->
+        {
+            // we need to keep tracing the cause to print out the real failure reason
```
nedleitch, line 25: I don't think we really need the comment :)

cluster/src/main/java/com/hartwig/pipeline/failsafe/DefaultBackoffPolicy.java
```diff
@@ -20,9 +20,16 @@ public class DefaultBackoffPolicy<R> extends RetryPolicy<R> {
...
+            Throwable lastFailure = rExecutionAttemptedEvent.getLastFailure();
+            while(lastFailure != null)
+            {
+                LOGGER.warn("[{}] failed: {}", taskName, lastFailure.getMessage());
```
nedleitch, line 29: How much are we actually going to print this way? Would it make sense to _only_ print the bottom exception and its stacktrace?

hongwingl: After the change, an error looks like this:
```
2024-05-10 12:48:26 WARN - [Lifecycle manager operation [insertVm]] failed: com.google.api.gax.rpc.InvalidArgumentException: Bad Request
2024-05-10 12:48:26 WARN - [Lifecycle manager operation [insertVm]] failed: Bad Request
2024-05-10 12:48:26 WARN - [Lifecycle manager operation [insertVm]] failed: 400 Bad Request
POST https://compute.googleapis.com:443/compute/v1/projects/hmf-crunch/zones/europe-west4-b/instances
{
  "error": {
    "code": 400,
    "message": "Number of local SSDs for an instance of type c2-standard-60 should be one of [0, 8], while [4] is requested.",
    "errors": [
      {
        "message": "Number of local SSDs for an instance of type c2-standard-60 should be one of [0, 8], while [4] is requested.",
        "domain": "global",
        "reason": "badRequest"
      }
    ]
  }
}
```

Before the change it only prints

```
2024-05-10 12:48:26 WARN - [Lifecycle manager operation [insertVm]] failed: com.google.api.gax.rpc.InvalidArgumentException: Bad Request
```

Perhaps the last one is enough? I am not sure.

nedleitch: Yeah that's what I'm getting at, we only really care about the "bottom" one. But I think that's an optimisation, I'm fine with it being left the way it is.

nedleitch (approved):

PR #574, opened 2024-06-24
DEV-4026: Prevent some out of memory crashes

DFKoetsier:
Gave Lilac VM more RAM and heap. Reduced core count of lane alignment VM.


DFKoetsier:
Tested this on a few samples for the 38 project. They succeeded without issues.

stefvanlieshout (approved):

PR #575, opened 2024-07-05
INFRA-595: upgrade log4j to v2

hongwingl:
Upgrade log4j to log4j2. Half of our projects use log4j2 already so upgrading helps avoid conflicts.
Hpoefully there is not reason not to upgrade?


pom.xml
```diff
@@ -256,8 +255,8 @@
             <artifactId>value</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
```
mdenburger, line 259: All classes depend on the `org.slf4j.Logger` class (as they should), so the slf4j-api dependency should remain in here too.

pom.xml
```diff
@@ -242,9 +236,14 @@
             </dependency>
 
             <dependency>
-                <groupId>org.slf4j</groupId>
-                <artifactId>slf4j-log4j12</artifactId>
-                <version>${slf4j-log4j12.version}</version>
+                <groupId>org.apache.logging.log4j</groupId>
+                <artifactId>log4j-core</artifactId>
+                <version>${apache.log4j.version}</version>
+            </dependency>
```
mdenburger, line 242: Why do we need a dependency on log4j-core? It should not be necessary, only log4j-slf4j-impl should be enough to get logging going.

mdenburger (changes requested):
Nice upgrade!

pom.xml
hongwingl: thanks you are right. Shows my lack of understanding of SLF4j. I will fix it.

mdenburger (approved):

PR #583, opened 2024-08-21
INFRA-696 Run SnpGenotype in HG38 mode

stefvanlieshout:
This change makes sure that for Targeted mode (which runs on HG38 in our production setup) also output the `snp_genotype.vcf` file with 26 or 31 SNP genotype calls. This will allow performing of snpcheck once the MIP-based snpcheck is ready for production.

Tested this change by running on TARGETED COLO from `gs://hmf-verification-data-fastq/COLO829-TARGETED/` and comparing the resulting snp genotype VCF with the same file from WGS COLO, which yielded a perfect match for all 26 positions.

### DIFF
--- a/cloudbuild-pr.yaml
+++ b/cloudbuild-pr.yaml
@@ -20,6 +20,7 @@ steps:
     args:
       - 'install'
       - '-Drelease'
+      - '--debug'
       - '--batch-mode'
     env:
       - MAVEN_OPTS=-Dmaven.repo.local=/cache/.m2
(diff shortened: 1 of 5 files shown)


nedleitch:
Stef was forced to make a new Docker image for pipeline5's build after I somewhat callously cleaned up the Dockerhub repository two weeks ago. Unfortunately this wasn't straightforward and the build failed with very little useful information.

I certainly hope my commits are squashed, the difference between Stef's final commit and my final one is actually quite small. I eventually managed to reproduce it locally and find the problem was to do with the Docker container's underlying operating system, so the changes here are actually quite minimal overall.

See the message on my final commit for a summary.

DFKoetsier (approved):
Looks good to me

PR #588, opened 2024-09-16
INFRA-786 Add 31SNP genotype output for MIP-based snpcheck

stefvanlieshout:
Also added a sample json to run COLO mini from BAM to speed up testing (when running pipeline5 from Intellij).


DFKoetsier (approved):
LGTM!

PR #590, opened 2024-09-24
INFRA-505: Always include RUO disclaimer on Orange report

DFKoetsier:
Apparently I hadn't merged this here yet.


DFKoetsier:
The smoke tests fail in the same locations as before this change. All other tests pass.

stefvanlieshout (approved):
Looks good, nice simplification also.

PR #596, opened 2024-10-18
DEV-4103: Add option to redo duplicate marking

DFKoetsier:
When providing a BAM or CRAM and the new command line argument `-redo_duplicate_marking` is set to true, alignment is skipped, but Redux is rerun on the BAM or CRAM.

We need this if we want to be able to correctly run SAGE on old BAMs or CRAMs without converting back to FASTQ first.

`-redo_duplicate_marking` defaults to `false`.


DFKoetsier:
I have tested COLO mini BAM and CRAM input with the new option set to true or false and I have compared it with a run from FASTQ with the new option set to false. The only difference between the resulting Orange JSONs is in the values set for the LPS, which aren't consistent between runs anyway.

BAM input with `-redo_duplicate_marking false`: gs://pipeline-output-dev/COLO829v003T-bam-testing
BAM input with `-redo_duplicate_marking true`: gs://pipeline-output-dev/COLO829v003T-dedup-redo-test
CRAM input with `-redo_duplicate_marking false`: gs://pipeline-output-dev/COLO829v003T-no-redo-cram-test
CRAM input with `-redo_duplicate_marking true`: gs://pipeline-output-dev/COLO829v003T-dedup-redo-cram-test
FASTQ input with `-redo_duplicate_marking false`: gs://pipeline-output-dev/COLO829v003T-testing

stefvanlieshout (approved):

PR #597, opened 2024-10-23
DEV-4108: Fix smoke tests

DFKoetsier:


stefvanlieshout (approved):
Assuming that the removal of `tumorOnlyCommands` for CUPPA was intended, all looks fine to me. And thanks for making smoke tests no longer fail :)

DFKoetsier:
> Assuming that the removal of `tumorOnlyCommands` for CUPPA was intended, all looks fine to me. And thanks for making smoke tests no longer fail :)

This was intended, yes. I confirmed with Luan that CUPPA is not supposed to run in tumor-only mode.

PR #604, opened 2025-02-24
INFRA-1249 Migrated cloudbuild to use new repository

kzagranovskyi-hw:


mvanniekerkHartwig (commented):
Please also make the same changes in the `AUS-243` branch.

PR #605, opened 2025-02-24
INFRA-1269 disable soft delete and set auto delete on run buckets

mvanniekerkHartwig:


mvanniekerkHartwig:
Addressed by disabling default soft delete on the project level.

PR #606, opened 2025-02-28
Infra 1285

mvanniekerkHartwig:
expect jitter params and ms table when running in skip markdups mode

### DIFF
--- a/cluster/src/main/java/com/hartwig/pipeline/alignment/bwa/BwaAligner.java
+++ b/cluster/src/main/java/com/hartwig/pipeline/alignment/bwa/BwaAligner.java
@@ -79,9 +79,9 @@ public class BwaAligner implements Aligner {
     private final ExecutorService executorService;
     private final Labels labels;
 
-    public BwaAligner(final Arguments arguments, final ComputeEngine computeEngine, final Storage storage, final PipelineInput input,
-            final SampleUpload sampleUpload, final ResultsDirectory resultsDirectory, final ExecutorService executorService,
-            final Labels labels) {
+    public BwaAligner(final Arguments arguments, final ComputeEngine computeEngine, final Storage storage,
+            final PipelineInput input, final SampleUpload sampleUpload, final ResultsDirectory resultsDirectory,
+            final ExecutorService executorService, final Labels labels) {
         this.arguments = arguments;
         this.computeEngine = computeEngine;
         this.storage = storage;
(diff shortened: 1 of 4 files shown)


cluster/src/main/java/com/hartwig/pipeline/alignment/bwa/BwaAligner.java
```diff
@@ -129,6 +142,43 @@ public AlignmentOutput run(final SingleSampleRunMetadata metadata) throws Except
...
+    private GoogleStorageLocation locateFile(SingleSampleRunMetadata metadata, String bamLocation,
+            com.hartwig.pipeline.reference.api.DataType dataType) {
+        // Because the tumor and reference are not nullable, but they are not necessary for location the file, we use the placeholder ___
+        var tumor = metadata.type().equals(SingleSampleRunMetadata.SampleType.TUMOR) ? metadata.sampleName() : "___";
+        var reference = metadata.type().equals(SingleSampleRunMetadata.SampleType.REFERENCE) ? metadata.sampleName() : "___";
+        var pipelineRun = new PipelineRun(Pipeline.DNA_6_0, tumor, reference);
+        var refDataSampleType = metadata.type().equals(SingleSampleRunMetadata.SampleType.TUMOR) ? SampleType.TUMOR : SampleType.REFERENCE;
+        // We assume the default output format for pipeline5, both the BAM and the CRAM file are 2 levels below the pipeline output root.
+        var pipelineOutputLocation =
+                new PipelineOutputTemporaryLocation(URI.create(bamLocation).resolve("../../"), PipelineOutputStructure.PIPELINE5);
```
DFKoetsier, line 162: When starting platinum runs our file structure for the input doesn't necessarily mimic a pipeline structure. The BAM or CRAM can simply be "in a bucket directory, somewhere". To what extent does this code require a specific directory structure? Like, does the directory need to be named after the reference or tumor ID or something.

Though it's not like that would have worked before, so this would still be an improvement. At least there would be a way to get it working now.

mvanniekerkHartwig: Indeed, it's magical but I did not see a better option. Long term PDL/platinum should provide these as extra arguments, but my goal was to update pipeline5 in a backwards compatible way.
The code chooses to assume these files are located in the same place as they are in the database relative to the bam file, and in the same place as in the pipeline5 output location. That to me seems to be choice that makes the least assumptions.

DFKoetsier (approved):
LGTM!

cluster/src/main/java/com/hartwig/pipeline/alignment/bwa/BwaAligner.java
DFKoetsier: Sounds like the right trade-off, indeed.

PR #610, opened 2025-04-10
Infra 1398 - fix running from cram using cram2bam when not rerunning redux

mvanniekerkHartwig:


DFKoetsier (approved):
The code looks good to me. Assuming the build succeeds, I would be happy with merging this.

mvanniekerkHartwig:
I ran one "real" sample already using this code and the cram2bam configuration, that one succeeded. Hopefully the smoke tests don't have surprises in store for us this time ;)

PR #620, opened 2025-05-07
DEV-4176: Write non-PASS Purple variants in targeted mode

DFKoetsier:


edk360 (approved):
No comments. Output checked for one test run and looks good.

PR #624, opened 2025-08-14
INFRA-1707 Add optional arguments to override #cpus for certain stages

mdenburger:
Very similar to overriding the memory for stages. I added a second regex argument to keep everything fully backwards compatible.


mvanniekerkHartwig (approved):
nice one!

